PR #13, opened 2017-02-13
New PID layout

si618:
Nice PID layout suggested by Vasily Sadvariy.


raphaelcoeffic:
@si618 I haven't tried it yet, but did you have a look at the proposed new layout for X7? It's in #14 and it looks really great. Do you think we could have the same for X9? Also, I think it would be better to rebase your change on top of master instead double-merging ;-)

si618:
Closed as @JR63 has included X9 in his PR, cheers! :)

PR #17, opened 2017-02-26
Pit mode switch

punkkills:
Hey, first off, forgive me if I am not doing this correctly.  While I'm a software engineer, I don't have any experience with git or it's workflow :D. 

This code lets you disable pit mode with a switch through telemetry.  It has some safety features built in.  It also does not require that the telemetry screen is visible.

You need to assign the pm_arm and pm_d-arm functions to different positions of a switch.  In order for pit mode to be deactivated, (armed) the function pm_d-arm first has to run, followed by pm_arm.   This allows some safety such that you can't just power up your quad with the video arming switch enabled and start transmitting, you must first first hit the disarmed state.

Next, I added a background task and refactored the telemetry communication stuff to that so the switch can be used with the telemetry screen closed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,4 @@ luac.out
 *.x86_64
 *.hex
 
+install.bat
--- a/Horus.lua
+++ b/Horus.lua
@@ -70,7 +70,7 @@ backgroundFill = TEXT_BGCOLOR
 foregroundColor = LINE_COLOR
 globalTextOptions = TEXT_COLOR
 
-local run_ui = assert(loadScript("/SCRIPTS/BF/ui.lua"))()
+local run_ui, background_ui = assert(loadScript("/SCRIPTS/BF/ui.lua"))()
 
 drawScreenTitle = function (title)
   lcd.drawFilledRectangle(0, 0, LCD_W, 30, TITLE_BGCOLOR)
@@ -78,4 +78,4 @@ drawScreenTitle = function (title)
   --lcd.drawText(LCD_W-40, 5, page.."/"..pages, MENU_TITLE_COLOR)
 end
 
-return {run=run_ui}
+return {run=run_ui, background=background_ui}
--- a/X7.lua
+++ b/X7.lua
@@ -69,5 +69,5 @@ MenuBox = { x=1, y=10, w=100, x_offset=36, h_line=10, h_offset=3 }
 SaveBox = { x=20, y=12, w=100, x_offset=4,  h=30, h_offset=5 }
 NoTelem = { 36, 55, "No Telemetry", BLINK }
 
-local run_ui = assert(loadScript("/SCRIPTS/BF/ui.lua"))()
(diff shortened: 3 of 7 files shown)


ronerone:
@punkkills Hi!
This is a great improvement that i would have requested.
It will help some pilots during preflight steps in races, but moreover, it will help begginers as i am save their tramp from overheating before flying.
Thank you for that, i'm looking forward to getting it in new release.

Ronan.

raphaelcoeffic:
@punkkills Thx a lot! That opens quite some use-cases! However, as I wrote you, I see that more as a work-around for now, as BF is going to have support for switches for these kind of changes rather soon. So I'd prefer leave this one as a PR for now. Please tell me your thoughts.

mikeller:
@raphaelcoeffic: I think support for switches might be some way out yet, since there is some work needed on the code that handles switches, because we're just about to max out on the number of switches that the current code can handle.
Also, in practical terms, I think the usefulness of switches is far more limited than the usefulness of MSP over SmartPort. After all, switches tie up the limited resource of channels, so a lot of users will probably have to remove another switch to free up a channel in order to fit in a new switch.

punkkills:
@raphaelcoeffic I'll leave it up to you.  As @mikeller mentioned, I was interested in the prospect of not needing another channel in BF (depending on if it was implemented as a mode or an adjustment).  In my opinion (biased :D)  it adds a lot of value right now and maybe into the future if the BF feature is a ways off.  If you aren't interested in pulling it, I would probably put it out there for people to use but wouldn't be excited about diverging from what you have put together.

raphaelcoeffic:
@mikeller so what you're saying is that it will take longer that I think, and at the end, it might not be as useful as I think it could?

mikeller:
@raphaelcoeffic At the moment there is only space for one more switch in the 32 bit set that is used for switches, so adding the pit mode switch will be possible quickly, but that's it. As to usefulness, I think seen on its own it will be very useful, but since the number of switches that can be reasonably used is limited by the number of RX channels, using it through MSP is going to be more useful for users that do not have any more unused RX channels available.

raphaelcoeffic:
@punkkills just a quick question: are you using a SmartAudio VTX? I'm just wondering because there is no way to power up the Tramp in pit-mode for now. If this is the case, do you set your TBS unify into race-mode by pressing the button while powering up? (at least the first time)

punkkills:
Yes, I'm using a Unify Race.  I didn't realize the tramp couldn't be powered up in pit mode.  On the Unify, you put it into pit mode by pressing the button on power up like you say.  When you disable pit mode with the switch, it is temporary and will boot in pit mode on next power cycle.  You have to boot again with the button to make it transmit on startup.

mikeller:
This repository has been rebased on the more advanced https://github.com/codecae/betaflight-tx-lua-revised. Please update your pull request if it is still relevant, or close it otherwise.

PR #58, opened 2017-11-13
Updating to new file structure

codecae:
Wholesale replacement of the lua scripts.  Reworked file system structure to enable contextual dynamic loading of radio and protocol specific code.  Common libraries have been created (ui, msp) so very few, if any, changes need to be performed in the core codebase when making adding or adjusting screen code (adding new transmitters, layouts, etc).

Tested on both X9 and X7 with success.

Would appreciate others to also test this code to confirm success.  The most recent build can be downloaded from: http://null.eastus.cloudapp.azure.com/job/Betaflight%20Lua%20-%20X7,%20X9,%20X9D,%20X9D+/

This download includes precompiled binaries.


raphaelcoeffic:
@codecae I fixed the basics in the Horus part, but we're far from a proper layout there... Still need a bit of time. However, I noticed something else that might apply to the other TXs as well: the field tab order is bit weird, jumping back and forth across the page. I think this needs fixing (or is that only in the horus script?).

Although, I'm not sure we have all the latest functional changes included here. Please have a look at the commits from the last 2 month.

PR #57, opened 2017-11-13
Make vtx freq editable (rebased)

ethomas997:
This adds the functionality in PR #41, rebased to the current 'master'.

Can step through available frequencies for bands / channels.  Implements feature suggested by issue #33.

If Cleanflight/Betaflight has 'SetFreqByMHzMsp' support then any frequency in MHz can be set when the band is set to 'U'.  (The channel field may be used to set the frequency in 100-MHz increments.)  The current Betaflight 'master' supports this functionality.

This also includes the changes in PR #56 . (Note save-page issues described in that PR.)

For 'SetFreqByMHzMsp' documentation, see:  http://www.etheli.com/CF/vtxCfgCMSFreqViaMsp

--ET

PR #62, opened 2017-11-13
draft horus version

raphaelcoeffic:
This adds supports for the Horus X10(S) and X12S to the new refactored scripts.

A couple of items are not 100% finished:
 - [ ] menu box glitch
 - [ ] "short-cut" out of TELEMETRY directory to allow easy one-time execution (Telemetry scripts do not exist as such on the Horus).

Tests / TODO:
 - rate2 and filters were not working with my quad, which does not mean anything, since I'm running an old (from last year) BF.

### DIFF
--- a/src/SCRIPTS/BF/X12S/filters.lua
+++ b/src/SCRIPTS/BF/X12S/filters.lua
@@ -1,0 +1,45 @@
+
+return {
+    read              = 92, -- MSP_FILTER_CONFIG
+    write             = 93, -- MSP_SET_FILTER_CONFIG
+    eepromWrite       = true,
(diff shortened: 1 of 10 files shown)


putimir:
Hi, what about the x10/12 layouts from the revised branch, that were added [#51](https://github.com/codecae/betaflight-tx-lua-revised/pull/51) - [#57](https://github.com/codecae/betaflight-tx-lua-revised/pull/57)?

Also save menu can would look better like [#50](https://github.com/codecae/betaflight-tx-lua-revised/pull/50), but i did'nt know where to put it since the code is changin too rapidly for me :)

PR #66, opened 2017-11-14
Updated radios to include low resolution settings

codecae:
Added lcdResolution.low designation for low-res radios (x7 and x9)


putimir:
Hi, what about x12s-simu and x10-simu as supported radio?

codecae:
@putimir If you want to open an issue to start the dialog or submit a pull request, we can look into adding those radios.

codecae:
May make sense to do simu for all

PR #73, opened 2017-11-14
Corrected errors occuring during fast PWM and VTX page switches.

codecae:
When pages are switched quickly on SmartPort, there are arithmetic issues in the page update code for PWM and VTX screens. This change corrects the failures when calculating pid rates and vtx frequency.


raphaelcoeffic:
@null shouldn't we actually have all these functions centrally, instead of duplicating them all over the place? I know started with this mess, but you seem enjoying correcting these things ;-)

codecae:
Totally agree with that.. was going to propose it but wanted the dust to settle a bit.

codecae:
I like dry code :)  I can take a stab at it.

PR #77, opened 2017-11-14
simplified increment functions (page, line, etc…)

raphaelcoeffic:
Thanks to @DieHertz for the nice hint with the modulo + base for negative increments!

### DIFF
--- a/src/SCRIPTS/BF/ui.lua
+++ b/src/SCRIPTS/BF/ui.lua
@@ -133,22 +133,22 @@ local function processMspReply(cmd,rx_buf)
     end
 end
 
-local function incMax(val,inc,base)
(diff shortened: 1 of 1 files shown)


src/SCRIPTS/BF/ui.lua
```diff
@@ -133,38 +133,22 @@ local function processMspReply(cmd,rx_buf)
     end
 end
 
-local function MaxLines()
-    return #(Page.fields)
+local function incMax(val,inc,base)
```
mikeller: Using space after comma (like in natural language) makes it a bit easier to read.

raphaelcoeffic: and here are the spaces ;-)

PR #79, opened 2017-11-15
"Soft" reverting the background function

codecae:
Using a comment to "soft" revert the background function. I, as well as a few other users, have found that it seems to be preventing access to the telemetry buffer.

### DIFF
--- a/src/SCRIPTS/TELEMETRY/bf.lua
+++ b/src/SCRIPTS/TELEMETRY/bf.lua
@@ -26,4 +26,5 @@ function run_bg()
   end
 end
 
-return { init=background.init, run=run, background=run_bg }
+--return { init=background.init, run=run, background=run_bg }
+return { run=run_ui }


codecae:
@mikeller @raphaelcoeffic 

Exhibited behavior:
The background function appears to be hogging the telemetry buffer.  It begins running when the model is loaded and immediately prevents script access to the telemetry buffer.  Both one-time and telemetry scripts such as bf.lua are affected.  Consequently, no msp or crsf requests for device configuration are being sent through the transport layer.

Steps to reproduce on an x9d+:

1) disable the 'bf' telemetry screen
2) build and then copy the scripts from commit 2443fbf to the SD.  _Don't enable_ the 'bf' telemetry screen right away when restarting.
2a) if you have CRSF, you can confirm that the telemetry buffer is available by running the crossfire.lua script before configuring the 'bf' telemetry screen.
3) enable the 'bf' telemetry screen in opentx
4) Hold page to open the telemetry screen and no values will appear on any of the pages.
4b) if you have CRSF, attempting to run the crossfire.lua at this point results it "Waiting for Crossfire devices..."
5) disable the 'bf' telemetry screen and boot into bootloader mode.
6) build and copy the scripts from commit 21ffcd2 to the SD. _Don't enable_ the 'bf' telemetry screen right away after restarting.
6b) if you have CRSF, run the crossfire.lua script and you will see that the buffer is available
7) enable the 'bf' telemetry screen in opentx
8) invoke the telemetry screen and you will now see that values are available.
8b) if you have CRSF, run the crossfire.lua script and you will see that it no longer says "Waiting for Crossfire devices..."

mikeller:
Unable to reproduce this here with OpenTx 2.2, *unless* the `crossfire.lua` script is run, so it looks like it is the `crossfire.lua` script that is causing this.

codecae:
I can recreate the issue without ever running crossfire.lua, as well.

raphaelcoeffic:
@mikeller @codecae I just looked up something in the opentx code, and it seems that a single call to `crossfireTelemetryPush` can block the telemetry output buffer (there is only one shared by crossfire & smartport) for ever, IF the Tx is not using a crossfire module.

basdelfos:
Can not reproduce this also (don't have the crossfire.lua running). Have been using the bf lua scripts with background process running without any issues.

PR #81, opened 2017-11-15
Made rssi and datetime bg function portable to crsf and sport

codecae:
Used protocol code to make the rssi and datetime updates protocol agnostic.


src/SCRIPTS/BF/background.lua
```diff
@@ -1,48 +1,26 @@
...
-
         -- ------------------------------------
         -- SYNC DATE AND TIME
         -- ------------------------------------
-
-        -- get sensor value
-        local newSensorValue = getValue(sensorId)
-
-        if not timeIsSet and type(newSensorValue) == "number" and newSensorValue > 0 then
+        if not timeIsSet and protocol.push() then
```
mikeller: This doesn't work (at least on SmartPort), since the telemetry connection is never actually lost when the flight controller is rebooted but the RX isn't power cycled.

raphaelcoeffic:
Looks like we're good now!

mikeller:
@raphaelcoeffic: Why was this merged? It looks like you did not even test if what I remarked in my review is fixed, since it clearly is not.

raphaelcoeffic:
It has been, as it seems, and the new code was exactly what you were asking for. The link says "Show outdated" / "Hide outdated".

mikeller:
Huh? What are you talking about? Do you even read review comments before you click merge?

raphaelcoeffic:
@mikeller of course!!!

mikeller:
So why did you merge it then without even checking if the ''This doesn't work' comment that I made did still apply?

raphaelcoeffic:
@mikeller I checked the updated code. It's now using the proper sensors. Look for yourself!

mikeller:
@raphaelcoeffic: Please read my comment above: It does not work. You merged code that is still broken in a way that is reproducible by my description above. Please don't do that.

raphaelcoeffic:
@mikeller please accept my apologies, I was sure I had understood the issue, which I clearly didn't.

mikeller:
All good now, we've discussed it, let's move on and try to improve processes for this repository in the future.

codecae:
Crossfire MSP_SET_RTC:
image

Crossfire MSP_TX_INFO:
image

codecae:
SmartPort MSP_SET_RTC:
image
image

Smartport MSP_TX_INFO:
image

raphaelcoeffic:
@codecae ok, so you do see the packets being sent, that's good. But the time is still not set, or is it?

codecae:
That was next :)

codecae:
I'm just supplying some artifacts that the output is being pushed to the telem buffer and is hitting the rx

raphaelcoeffic:
@codecae wait a sec: for the SPORT MSP_SET_RTC, don't you see the second packet? There should be a second one. The second one you posted is the reply from the FC, not request - part2.

codecae:
it's only one byte in length.  padded with a zero then crc.

raphaelcoeffic:
ok, so the second packet IS there, that good as well.

codecae:
wait.. I may have grabbed the wrong stuff.. there's something not right about the smartport one.. redoing it now

codecae:
Ok those look better now.  I don't see that it's updating, though.  Will need to run a debugger on the bf side, because it looks like the data is being sent.

codecae:
works for smartport

raphaelcoeffic:
> Ok those look better now. I don't see that it's updating, though. Will need to run a debugger on the bf side, because it looks like the data is being sent.

Same issue here... I'm using HEAD of master, and the time is NOT set, whereby I can see the requests as well.

codecae:
image

codecae:
looks like a timing thing... maybe need to use a different sensor.  If I power up the fc, then the radio, it works.  Other way around, it does not.

raphaelcoeffic:
I removed the telemetry script to test with a one-time only, and I cannot get it to work....

```
SCRIPT_HOME = "/SCRIPTS/BF"

protocol = assert(loadScript(SCRIPT_HOME.."/protocols.lua"))()
radio = assert(loadScript(SCRIPT_HOME.."/radios.lua"))()

assert(loadScript(radio.preLoad))()
assert(loadScript(protocol.transport))()
assert(loadScript(SCRIPT_HOME.."/MSP/common.lua"))()

local lastReqTS   = 0
local INTERVAL    = 50
local MSP_SET_RTC = 246
local reqSent     = 0

local function checkTime()
   if getTime() - lastReqTS > INTERVAL then
      local now = getDateTime()
      local year = now.year;

      values = {}
      values[1] = bit32.band(year, 0xFF)
      year = bit32.rshift(year, 8)
      values[2] = bit32.band(year, 0xFF)
      values[3] = now.mon
      values[4] = now.day
      values[5] = now.hour
      values[6] = now.min
      values[7] = now.sec

      -- send msp message
      if protocol.mspWrite(MSP_SET_RTC, values) then
         reqSent = reqSent + 1
      end
   end
    -- process send queue
   mspProcessTxQ()
end

local function run()

   checkTime()

   lcd.clear()
   lcd.drawText(15, 30, "reqSent:", 0)
   lcd.drawNumber(88, 30, reqSent, 0)

   return 0
end

return { run=run }
```

codecae:
it works sporadically.  trying to find a way to reproduce reliably.  I've seen the clock get set on both formats.  something about the timing is off.

raphaelcoeffic:
where? on the BF side?

PR #85, opened 2017-11-16
Updated build script to fail upon compilation error and report status of all files

codecae:
Current script will report build failures but exits 0 in CI when failures exist.  This update will provide all file compilation statuses and exit on failure if one of them has a build issue.

### DIFF
--- a/bin/build.sh
+++ b/bin/build.sh
@@ -9,6 +9,7 @@ fi
 cp -fR src/* obj
 
 MANIFEST=(`find obj/ -name *.lua -type f`);
+LAST_FAILURE=0
 
 if [ ${#MANIFEST[@]} -eq 0 ]; then
     echo -e "\e[1m\e[39m[\e[31mTEST FAILED\e[39m]\e[21m No scripts could be found!."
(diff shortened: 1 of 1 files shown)


bin/build.sh
```diff
@@ -20,9 +21,15 @@ do
...
+        FAIL=$((FAIL | _fail))
+        echo -e "\e[1m\e[39m[\e[31mBUILD FAILED\e[39m]\e[21m Compilation error in file ${SRC_NAME}\e[1m"
+    fi
 done
 
-echo -e "\e[1m\e[39m[\e[32mTEST SUCCESSFUL\e[39m]\e[21m All lua files built successfully!"
+if [[ $FAIL -eq 0 ]]; then
+    echo -e "\e[1m\e[39m[\e[32mTEST SUCCESSFUL\e[39m]\e[21m All lua files built successfully!"
+fi
+exit $FAIL
```
mikeller: Exiting with a status that is a logical or of all exit codes is a bit awkward, and makes it impossible to do something meaningful with the exit status. We should probably rather exit with the last failed compilation exit status.

codecae:
These changes looking any better?

mikeller:
@codecae: Much better. :-)

PR #88, opened 2017-11-17
Updated link to direct users to the releases page for downloads

codecae:


README.md
```diff
@@ -2,7 +2,7 @@
 
 ### How to download the scripts
 
-[Click here](http://null.eastus.cloudapp.azure.com/job/Betaflight%20Lua%20-%20X7,%20X9,%20X9D,%20X9D+/lastSuccessfulBuild/artifact/obj/SCRIPTS/*zip*/SCRIPTS.zip) to download a zip file containing latest version.
+[Click here](http://null.eastus.cloudapp.azure.com/job/Betaflight%20Lua%20-%20X7,%20X9,%20X9D,%20X9D+/lastSuccessfulBuild/artifact/obj/*zip*/obj.zip) to download a zip file containing latest version.
```
mikeller: We *really* need to get away from asking users to install the latest nightly, and start doing proper releases. So instead of putting more lipstick onto this pig, I suggest we take what we've got now, and make release 0.1 from it!

codecae: Absolutely... I was hoping by now we'd be on ci.betaflight.tech, but I will make the changes there.  I'm having to push these changes to my fork in order to update Jenkins, which is kind of kludgy.

Should we tag the release in the correct repo and I'll create the release from that?

mikeller: That's where the release should go, and where all of the documentation should be pointing to: https://github.com/betaflight/betaflight-tx-lua-scripts/releases

If we are not ready yet to use http://ci.betaflight.tech/ to do release builds, using your Jenkins on azure is still better than using a local build, and I think right now, getting a release out and starting to get users used to using releases instead of latest builds is better than keeping them pulling from Jenkins, and getting all upset if something is broken.

codecae: v1.0 has been released based on #87.  I have updated the commentary on jenkins, as well.

mikeller: Now all that is left to be done is to update the readme to tell the users to go and get the release. :smiley:

codecae:
The latest amend has an update for the readme

PR #91, opened 2017-11-19
Removed references to undeclared 'mspOutOfOrder' variable

ethomas997:
Would throw an error if the out-of-order condition happened to occur.

--ET


raphaelcoeffic:
Very good catch! That an ancient relief of my first tests.

PR #41, opened 2017-09-25
Made vtx freq editable

ethomas997:
(Note updated PR #92)

Can step through available frequencies for bands / channels.  Implements feature suggested by issue #33.

If Cleanflight/Betaflight has 'SetFreqByMHzMsp' support then any frequency in MHz can be set when the band is set to 'U'.  (The channel field may be used to set the frequency in 100-MHz increments.)

For 'SetFreqByMHzMsp' test builds and documentation, see:  http://www.etheli.com/CF/vtxCfgCMSFreqViaMsp

Betaflght PR:  https://github.com/betaflight/betaflight/pull/4265

--ET

Issue #33, opened 2017-06-27
[Feature] Ability to step through frequencies


zandzpider:
I would like to step through the frequencies in addition to selecting band and channel.

selecting a frequency would automatically select the band and channel for you.

find it annoying to figure out where "5800" is by selecting band and channel

robert-hoffmann:
+1

PR #92, opened 2017-11-20
Make vtx freq editable - revised

ethomas997:
This modifies the VTX script to support editing the 'frequency' field, so it will step through the available frequencies for bands / channels.  Implements the feature suggested by issue #33.  This is a rebase of PR #41.

Also, if Cleanflight/Betaflight has 'SetFreqByMHzMsp' support then any frequency in MHz can be set when the band is set to 'U'. (The channel field may be used to set the frequency in 100-MHz increments.)

 I've posted build files for a devel/beta build (https://github.com/betaflight/betaflight/commit/a9105e88c4be2bdeca1e5f66ae99f104004934e6) of Betaflight with 'SetFreqByMHzMsp' support, here:  http://www.etheli.com/CF/vtxCfgCMSFreqViaMsp/bf_builds_20171118_372

I've developed and tested this on a Taranis X7; testing on the other transmitter types would be good.

I've tested with Betaflight v3.2.2 and with the 2017-11-18 snapshot of Betaflight mentioned above.

The 'vtx.lua' files for X7/X9/Horus are identical except for the 'text' and 'fields' definitions, so most of the content could be broken out into a 'vtx_common.lua' file.

--ET

### DIFF
--- a/src/SCRIPTS/BF/HORUS/vtx.lua
+++ b/src/SCRIPTS/BF/HORUS/vtx.lua
@@ -4,10 +4,13 @@ return {
     write          = 89, -- MSP_VTX_SET_CONFIG
     eepromWrite    = true,
     reboot         = false,
-    saveMaxRetries = 2,
-    saveTimeout    = 300, -- 3s
     title          = "VTX",
     minBytes       = 5,
+    prevBandVal    = 0,
+    prevChanVal    = 0,
+    prevFreqVal    = 0,
+    lastFreqUpdTS  = 0,
+    freqModCounter = 0,
     text= {
         { t = "Band",    x = 36,  y = 110 },
         { t = "Channel", x = 36,  y = 155 },
@@ -18,12 +21,12 @@ return {
     },
     fields = {
         -- Band
-        { x = 130,  y = 110, min=1, max=5, vals = { 2 },  to = MIDSIZE,
-          table = { "A", "B", "E", "F", "R" },
-          upd = function(self) self.updateVTXFreq(self) end },
+        { x = 130,  y = 110, min=0, max=5, vals = { 2 },  to = MIDSIZE,
+          table = { [0]="U", "A", "B", "E", "F", "R" },
+          upd = function(self) self.handleBandChanUpdate(self) end },
         -- Channel
         { x = 130,  y = 155, min=1, max=8, vals = { 3 },  to = MIDSIZE,
-          upd =  function(self) self.updateVTXFreq(self) end },
+          upd =  function(self) self.handleBandChanUpdate(self) end },
         -- Power
         { x = 350,  y = 110, min=1, vals = { 4 },         to = MIDSIZE,
           upd = function(self) self.updatePowerTable(self) end },
@@ -33,9 +36,10 @@ return {
         -- Proto
         { x = 130,  y =  68, vals = { 1 },                to = MIDSIZE,
           write = false, ro = true,
-          table = {[3]="SmartAudio",[4]="Tramp",[255]="None"} },
+          table = { [1]="RTC6705",[3]="SmartAudio",[4]="Tramp",[255]="None"} },
         -- Freq
-        { x = 350,  y = 68, min=5000, max=6000, ro=true, to = MIDSIZE },
+        { x = 350,  y = 68, min = 5000, max = 5999, vals = { 6 }, to = MIDSIZE,
+          upd = function(self) self.handleFreqValUpdate(self) end },
     },
     freqLookup = {
         { 5865, 5845, 5825, 5805, 5785, 5765, 5745, 5725 }, -- Boscam A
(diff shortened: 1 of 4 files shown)


mikeller:
> I've tested with Betaflight v3.2.2 and with the 2017-11-18 snapshot of Betaflight mentioned above.

So this works with the released version of betaflight 3.2.2?

ethomas997:
@mikeller:  With BF v3.2.2, setting via band/channel should work the same as with previous versions of the VTX script.  The PR adds the ability to set band/channel by editing the value of the 'frequency' field -- as it is modified, the band/channel change to match the value.

If a recent devel / 'master' version of Betaflight is used, a band value of 'U' may be set and then any frequency value in MHz can be dialed in and set.

mikeller:
Ah I see. How is the case of RTC6705 handled? This driver does not support direct frequency entry.

ethomas997:
@mikeller:  I don't think any version of the VTX script has supported RTC6705.  Only SA and Tramp.

mikeller:
@ethomas997: Oh, why is that? I thought we moved to unified drivers for all 3 on the firmware side? @codecae?

codecae:
@mikeller it doesn't appear that the 6705 registers with a vtable with vtx common.  Don't think it ever has.

codecae:
The vtable is registered with vtx common on the firmware side, but it doesn't share the same feature set.

raphaelcoeffic:
I guess the VTX type is not displayed properly, but that should be all.

codecae:
I would assume this as well, but do not have hardware to verify.

raphaelcoeffic:
neither do I. Could anyone try this out?

src/SCRIPTS/BF/HORUS/vtx.lua
```diff
@@ -45,19 +49,150 @@ return {
         { 5658, 5695, 5732, 5769, 5806, 5843, 5880, 5917 }, -- RaceBand
     },
     postLoad = function (self)
-        if self.values[2] == 0 or self.values[3] == 0 or self.values[4] == 0 then
+        if (self.values[2] or 0) < 0 or (self.values[3] or 0) == 0 or (self.values[4] or 0) == 0 then
             self.values = {}
+        else
+            prevBandVal = 0                      -- reset value trackers
```
codecae: Scoping should be addressed here  These are specified as members but are being stored in the global area.

codecae: There are many other references below that should probably also be addressed.

src/SCRIPTS/BF/HORUS/vtx.lua
```diff
@@ -45,19 +49,150 @@ return {
...
+                rFreq = 0
+            end
+            if (self.values[2] or 0) > 0 then    -- band != 0
+                if self.values[3] and self.values[3] > 0 then
+                    if rFreq > 0 then             
+                        prevFreqVal = rFreq
+                        prevBandVal = self.values[2]
+                        prevChanVal = self.values[3]
+                        self.fields[1].min = 0   -- make sure 'U' band allowed
+                        eepromWrite = true
```
codecae: Scope should also be addressed on eepromWrite.  As a member, it should be accessed using self.

codecae: Does this provide a measurable benefit?  Seems like the write is fairly benign on older versions (even though the action isn't actually saving anything new).

ethomas997:
@codecae:  Is there a "philosophy" behind when to use "self.fields[x].value" vs "self.values[x]"?

codecae:
I don't think there's a philosophy, persay.  Generally speaking, I think the intent behind 'fields' is to give more free-form flexability to place things in locations that are unbound to a value cursor, where specifying configurations on values will bind the heading with the cursor and display them together.

mikeller:
@codecae, @raphaelcoeffic: I have a board with an onboard VTX, I can test if the scripts work with the RTC6705.

raphaelcoeffic:
> Is there a "philosophy" behind when to use "self.fields[x].value" vs "self.values[x]"?

@ethomas997 at the beginning, only `Page.values[x]` was used, as this is where the data are loaded when coming from the FC, and where they are taken when sending back to the FC. The values in the `fields` are basically derivates to be displayed.

src/SCRIPTS/BF/HORUS/vtx.lua
```diff
@@ -35,7 +38,8 @@ return {
           write = false, ro = true,
           table = {[3]="SmartAudio",[4]="Tramp",[255]="None"} },
         -- Freq
-        { x = 350,  y = 68, min=5000, max=6000, ro=true, to = MIDSIZE },
+        { x = 350,  y = 68, min = 5000, max = 5999, vals = { 6 }, to = MIDSIZE
```
mikeller: Missing comma.

mikeller:
The good news:

VTX control works with this pull request (and without it) for flight controllers with a VTX6705!

The bad news:

There's a syntax error. See comment.

ethomas997:
I'll fix that syntax error and address the other issues.  Also, should be able to show the RTC6705 device type.

mikeller:
That'd be great. All I get now is a lousy number:

img_20171121_234716

ethomas997:
I've implemented fixes; should be in good shape now:

Fixed syntax error in BF/HORUS/vtx.lua
Put in proper 'self.' scoping prefixes
Added 'RTC6705' to VTX Dev table
Added RTC6705 entry to VTX power table
Don't display Pit field if device is RTC6705

mikeller:
This looks a lot better now, thanks @ethomas997. The one thing that I am slightly concerned is that this adds quite a lot of code just to implement a relatively minor feature.

We are already getting problem reports (https://github.com/betaflight/betaflight-tx-lua-scripts/issues/97#issuecomment-346173273) from users trying to use the betaflight-tx-lua scripts together with other scripts, and this will make it worse.

What's your thoughts on this, @raphaelcoeffic, @codecae?

ethomas997:
@mikeller:  One significant space savings we could get right away would be to combine the 'vtx.lua' scripts for the three transmitter types.  The files are duplicates except for the 'text' and 'fields' entries.  If we break those out, there can be a single 'vtx.lua' script.

mikeller:
@ethomas997: That is a good point. We should probably do that before this is merged.

ethomas997:
@mikeller:  I was thinking we should it after this one is merged, as a separate PR, to better track the changes.

mikeller:
@ethomas997: Definitely in a separate pull request. I was thinking before, to avoid getting bug reports because of running out of memory, but if the consolidation is happening reasonably soon after this has been merged that will be fine as well.

ethomas997:
@mikeller I can make a point of doing the consolidation soon.

codecae:
I don't think that the space savings will make much of a difference regarding memory consumption.  Every time the screen is changed, garbage is collected and a new page object is loaded in place of the previous.  Pulling common methods from another file probably won't help this very much, but it would certainly help simplify maintenance of the code.

codecae:
The best way to ensure we are making use of our very precious memory resource is to be a good steward of variables and trim things down to the bare essentials.

raphaelcoeffic:
Exactly. The second thing would be to see whether we could split screens to make them smaller in terms of memory (MSP replies need to become  smaller).

mikeller:
@codecae: Sorry, my previous answer was given without studying the code in detail. I assumed that the files in question were all used on the same page, and not on separate pages. If they are separate pages, then the 'space savings' argument is largely void. But as you say, the cleanup aspect still persists.

ethomas997:
So, can this PR be merged in?

mikeller:
Sorry, it dropped off the radar. I am good with it, @codecae, @raphaelcoeffic, any objections?

ethomas997:
Could a new release be posted to include these changes?

mikeller:
Sorry we've been slacking off a bit. (Or rather spending time on other areas of Betaflight.)

Unfortunately now seems to be a bad time for a release of the scripts because of the memory shortage that 2.2.1 brought about - if we release now, the additional space requirements for the new features will likely push more users over the edge. I think the right course of action is to wait and see what happens to opentx/opentx#5579 (my hope would be for a bugfix release to improve the memory situation).

Again, sorry for not looking into this earlier.

ethomas997:
Understood.  You might want to add a note to the main README about the issues with OpenTX 2.2.1

mikeller:
Good idea, done.

PR #111, opened 2018-01-27
ESC telemetry page.

pulquero:
Added ESC (read-only) telemetry page.

### DIFF
--- a/src/SCRIPTS/BF/X7/x7pre.lua
+++ b/src/SCRIPTS/BF/X7/x7pre.lua
@@ -8,7 +8,8 @@ PageFiles =
     "filters1.lua",
     "filters2.lua",
     "pwm.lua",
-    "vtx.lua"
+    "vtx.lua",
+    "esc.lua",
 }
 
 MenuBox = { x=15, y=12, w=100, x_offset=36, h_line=8, h_offset=3 }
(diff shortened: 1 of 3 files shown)


src/SCRIPTS/BF/X7/x7pre.lua
```diff
@@ -8,7 +8,8 @@ PageFiles =
     "filters1.lua",
     "filters2.lua",
     "pwm.lua",
-    "vtx.lua"
+    "vtx.lua",
+    "esc.lua",
```
mikeller: Shouldn't this go into `X9/` and `HORUS/` as well?

codecae:
Given that this is a read-only extension, would it be better served up as a separate telemetry script?

pulquero:
The limit on the total number of scripts made me think I should be conservative in this respect.

src/SCRIPTS/BF/X7/x7pre.lua
pulquero: Sure, once you are happy with my approach I'll roll it out to the other radios too.

mikeller:
@codecae: How easy is it to do this as a separate telemetry script, but reuse the existing MSP functionality? Copying / reimplementing these parts would be very inefficient.

codecae:
The common map components can be reused.  They were separated from the UI for this reason.  All that would be required is the creation of a new event loop

codecae:
Using the existing UI may end up saturating the signal due to constant polling.  It could be better controlled with a purpose built event loop.

pulquero:
Closing in favour of https://github.com/betaflight/betaflight/pull/5581.

PR #124, opened 2018-03-30
Removed Event Table to save on memory

codecae:
Removed events.lua to prevent userEvent table from being added to the heap.  Replaced calls to the userEvent table with the original constants in OpenTX.


codecae:
Rather than rebasing, I'm of the opinion that custom frequency selection in master should be reverted and optimized.  It is also part of this lingering memory issue and should be taken back to the drawing board.

ethomas997:
Looks like the memory situation should improve when an OpenTX version with opentx/opentx#5579 is released (but I've not been following the situation there closely).

codecae:
I mean no disrespect regarding the revert @ethomas997 ... likely a temporary measure in order to get the memory issues addressed until a more adequate solution can be provided by opentx.

I have tested the effect of these changes on 2.2.1 and have not had a single memory failure:

Try [this version](https://drive.google.com/uc?authuser=0&id=1K5bsbJPHgzx6awizgWDLcDjFRSdraSJ1&export=download) if you'd like to assist with testing.

codecae:
I rescind my comment -- I did get a memory error, but it only occured when switching pages too quickly.  If the page is permitted to load completely before switching, the error does not occur.

src/SCRIPTS/BF/HORUS/vtx.lua
```diff
@@ -36,10 +33,9 @@ return {
         -- Proto
         { x = 130,  y =  68, vals = { 1 },                to = MIDSIZE,
           write = false, ro = true,
-          table = { [1]="RTC6705",[3]="SmartAudio",[4]="Tramp",[255]="None"} },
```
mikeller: I think the `RTC6705` entry should be reinstated.

ethomas997:
@codecae No disrespect taken; I know it's kind of a messy situation.  Do we have a sense of if opentx/opentx#5579 will fix this and if it will get into a release soon?

mikeller:
opentx/opentx#5579 should fix it - there are custom builds out there with this patch on top of 2.2.1, and they seem to be fixing the memory woes. But then, 2.2.2 might also include other changes that use more memory, negating this effect.
Re release, the last date that I've heard was 'mid February'. This has long passed, and no indication of a new release yet...

ethomas997:
@mikeller Understood.  You might want to check if the scripts still overflow memory even with the rollback.

BTW I have a "basic" version posted based on the previous edition of the scripts, with the "VTX freq editable" changes and two other screens (only), here:  https://github.com/ethomas997/betaflight-tx-lua-basic/releases

PR #132, opened 2018-05-23
Add GPS_Rescue

David-VG:
Added 2 screens, one for the GPS_Rescue PIDS and an other one with the main parameters.

Tested the X9 version, but X7 and Horus only with the simulator, if you have one, please verify it works fine.
image1 5
image2 2


mikeller:
@David-VG: Can you rebase this please?

Issue #134, opened 2018-05-25
Frsky X-lite not supported


jafc76:
I just received the Frsky X-lite and after to install the LUA Scripts for Betaflight, I'm getting this error:

Script syntax error
BF/radios.lua:43
Radio not supported: xlit

I know that it's a new radio so waiting for the compatibility with it.

Thanks in advance.

raphaelcoeffic:
@jafc76 there are certain unknowns as-of the available navigation keys inside the script (being sorted out in OTX dev chat right now). Once this is done, I might make the necessary adjustments.

raphaelcoeffic:
@jafc76 could you please test the code changes I've just made? Please bare with me, I have no way to test anything...

jafc76:
raphaelcoeffic, Thank you very much!!! It works!!!

The small joystic is working like this. If I move the joystick to the right, the cursor is moving down to the next option inside of that screen, if I move to the left, the same but going up inside of the same screen.
If I move the joystick up, I go to the next page, from PID screen, to rates1, rates2, etc... but if I move the joystick down to go to the previous page, doesn't go back.


Anyway, at the moment, it works, maybe a little tricky how the joystick works but it's functional. Thank you very much again!!

raphaelcoeffic:
Can you please confirm if you can reach the menu (long-UP)?
“Previous page” with the down key can be done easily.

jafc76:
To go to the LUA screen I have to press a long-down.

Inside of the LUA screen if I make a long-up I get one screen with this message:

Menu: save page
             reload
             reboot

raphaelcoeffic:
@jafc76 glad to hear it works properly!

raphaelcoeffic:
@jafc76 I made a small modification to support "previous page". The keys have been slightly changed to fit that logic:
- next page: DOWN
- previous page: UP
- pop-up menu: long DOWN

Would you mind re-testing?

jafc76:
I just copied the files to test it and works like a charm!

long-down to go to the LUA script.

right to go down and left to go up inside of each screen.

down to go to the next page and up to the previous one.

long down pop-up menu

Don't worry about to test it, I really appreciate your time and work!!

Eveything is good now, but if you change to move down pressing down, up pressing up, next page pressing right, previous pressing left, will be perfect!! Only if you can, no pressure.

Thanks again!!!

raphaelcoeffic:
[...]

PR #135, opened 2018-05-25
added support for X-Lite

raphaelcoeffic:
First shot at implementing #134.
WARNING: this is completely untested, and only made based on OpenTX source code.


DieHertz:
I'll test once I get batteries

PR #159, opened 2018-08-23
Change setting of RTC to use ticks.

mikeller:
To go with betaflight/betaflight#4682. Due to the dependency on this it will only work with firmware 4.0 and up, unfortunately. I do not know how to best deal with this, except allowing the user to change it by un-/commenting the relevant sections of code - supporting multiple API versions like it is done in configurator seems to be overly complex for the lua scripts.

Fixes betaflight/betaflight#6504.

### DIFF
--- a/src/SCRIPTS/BF/background.lua
+++ b/src/SCRIPTS/BF/background.lua
@@ -42,6 +42,7 @@ local function run_bg()
             -- or when connection is restored after e.g. lipo refresh
 
             if API_VERSION < 1.041 then
+                -- format: year (16) / month (8) / day (8) / hour (8) / min (8) / sec (8)
                 local now = getDateTime()
                 local year = now.year;
 
@@ -55,16 +56,16 @@ local function run_bg()
                 values[6] = now.min
                 values[7] = now.sec
             else
+                -- format: seconds after the epoch (32) / milliseconds (16)
                 local now = getRtcTime()
 
                 values = {}
-                values[1] = bit32.band(now, 0xFF)
-                now = bit32.rshift(now, 8)
-                values[2] = bit32.band(now, 0xFF)
-                now = bit32.rshift(now, 8)
-                values[3] = bit32.band(now, 0xFF)
-                now = bit32.rshift(now, 8)
-                values[4] = bit32.band(now, 0xFF)
+
+                for i = 1, 4 do
+                    values[i] = bit32.band(now, 0xFF)
+                    now = bit32.rshift(now, 8)
+                end
+
                 values[5] = 0 -- we don't have milliseconds
                 values[6] = 0
             end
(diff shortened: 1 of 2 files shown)


SRWieZ:
I don't upgrade my smaller quads too often (because there working great and are tuned). So, I prefer to have a working lua script with all version of betaflight.

I'm not aware of the complexity to have backward compatibility code. I assume that we have the API version, so what's the problem ? Memory or something else ? :)

mikeller:
@SRWieZ: Good point. We don't per se _have_ the API version, we have to send an MSP command to the flight controller to get it to return it. This adds more complexity to the lua scripts, and will push us closer to running out of RAM on the TX again, so it's not something that should be done lightly.

SRWieZ:
I see. We don't have enough RAM but we have plenty of storage, right ? Can we add different bf.lua with version number like bf33.lua bf35.lua bf40.lua ? So we could assign bf33 on older quad telemetry screen and bf40 for our beta testing quad. 

Should be easy to maintain, or I am missing something ?

SRWieZ:
My first thought was to use different branches for different versions on github like vendor libraries I'm working with. 

I don't know what's the best solution. :)

src/SCRIPTS/BF/background.lua
```diff
@@ -40,18 +40,34 @@ local function run_bg()
...
-            values[1] = bit32.band(year, 0xFF)
-            year = bit32.rshift(year, 8)
-            values[2] = bit32.band(year, 0xFF)
-            values[3] = now.mon
-            values[4] = now.day
-            values[5] = now.hour
-            values[6] = now.min
-            values[7] = now.sec
+            if API_VERSION < 1.041 then
+                local now = getDateTime()
```
fujin: The only thing that comes to mind is instead of having the two branches in the if/else expr, we do something like
```
assert(loadScript(SCRIPT_HOME.."/rtc_"..API_VERSION..".lua"))()`
```

which contains the api-version specific section.

src/SCRIPTS/BF/background.lua
```diff
@@ -40,18 +40,34 @@ local function run_bg()
...
+                values[3] = now.mon
+                values[4] = now.day
+                values[5] = now.hour
+                values[6] = now.min
+                values[7] = now.sec
+            else
+                local now = getRtcTime()
+
+                values = {}
+                values[1] = bit32.band(now, 0xFF)
```
fujin: perhaps a loop 1..4?

``` lua
local now = getRtcTime()
values = {}

for i = 1,2,3,4 do
  values[i] = bit32.band(now, 0xFF)
  now = bit32.rshift(now, 8)
end

values[5] = 0
values[6] = 0
```

src/SCRIPTS/BF/background.lua
```diff
@@ -40,18 +40,34 @@ local function run_bg()
...
+                now = bit32.rshift(now, 8)
+                values[2] = bit32.band(now, 0xFF)
+                now = bit32.rshift(now, 8)
+                values[3] = bit32.band(now, 0xFF)
+                now = bit32.rshift(now, 8)
+                values[4] = bit32.band(now, 0xFF)
+                values[5] = 0 -- we don't have milliseconds
```
fujin: Is it worth documenting in here the expected payload, or is it documented on the FW side re: MSP_SET_RTC?

PR #161, opened 2018-08-26
Disable VTX Power and Pit fields, when device type is unknown

milindur:
If the type of VTX is unknown, the power field (table) is not fully initialized. This results in an exception when unintentionally entering the power field.

This change disables power and pit fields, if the VTX device is unknown. Band and channel should remain functional, because the VTX could be connected to another device (PDB/OSD) (see https://github.com/betaflight/betaflight/pull/6522).


mikeller:
@milindur: Do you think this will fix #162?

milindur:
@mikeller I don't think so. In my case I got the error "BF/ui.lua:235; attempt to perform arithmetic on field 'max'. #162 seems to be something else.

PR #164, opened 2018-08-27
Fix for unable to update gyro lowpass. 

klutvott123:
When trying to update the gyro lowpass filter the value is set to it's original value when saving. 

This is because https://github.com/betaflight/betaflight/blob/793c5f7b71db746146a086e5c667759a0f13d415/src/main/interface/msp.c#L1248 contains TWO sbufWriteU8(dst, gyroConfig()->gyro_lowpass_hz);.

When sending the data from the lua script, the modified value is set here: https://github.com/betaflight/betaflight/blob/793c5f7b71db746146a086e5c667759a0f13d415/src/main/interface/msp.c#L1814

And the original value is set here: https://github.com/betaflight/betaflight/blob/793c5f7b71db746146a086e5c667759a0f13d415/src/main/interface/msp.c#L1833 overwriting the modified value. 

Sending only 18 bytes back prevents betaflight from entering that last if-statement. 

This is my first pull request. Hope i did it right.

### DIFF
--- a/src/SCRIPTS/BF/HORUS/filters.lua
+++ b/src/SCRIPTS/BF/HORUS/filters.lua
@@ -6,6 +6,7 @@ return {
     reboot            = false,
     title             = "Filters",
     minBytes          = 18,
+    outputBytes       = 18,
     text= {
         -- Column headers
         { t = "Gy LP",         x = 60,  y = 68, to = SMLSIZE },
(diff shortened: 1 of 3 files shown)


DieHertz:
It doesn't contain two, the second one is `sbudWriteU16`. The correct way would be to ignore the first byte and only the the U16 variant

mikeller:
@DieHertz: At the moment using the U16 is not possible without major code changes because of the way how lua pages are mapped to the contents of MSP requests. The solution proposed by @klutvott123 is in alignment with the fix for this that was implemented in #140.

PR #189, opened 2018-11-29
Increase smartport save timeout

klutvott123:
Increase smartport save timeout from 3 to 5 seconds. Increases the chances of saving on the first try with FPORT. Testing shows that very often it takes just a little over 3 seconds before the ``eepromWrite`` command is received back from the FC when using FPORT. If it doesn't arrive within 5 seconds, i think we can safely assume that it isn't coming at all and try again. 

The downside is that if it does fail to save, we have to wait a few seconds longer, but i think 5 seconds is a good compromise.

### DIFF
--- a/src/SCRIPTS/BF/protocols.lua
+++ b/src/SCRIPTS/BF/protocols.lua
@@ -10,7 +10,7 @@ local supportedProtocols =
         maxTxBufferSize = 6,
         maxRxBufferSize = 6,
         saveMaxRetries  = 2,
-        saveTimeout     = 300
+        saveTimeout     = 500
     },
     crsf =
     {


mikeller:
Is this tested and has it shown to improve the situation?

klutvott123:
Yes i have tested it. Here's some data that shows the time it takes between starting the save and the ``eepromWrite`` response is received:
```
3410ms
3650ms
4100ms
2450ms
3460ms
2560ms
3010ms
1710ms
2610ms
4900ms
1650ms
2060ms
3860ms
3110ms
4140ms
3200ms
4050ms
```
It still retries a lot, but at least now it has an increased chance of success. It's probably because of uart buffer overrun when using FPORT like we talked about a while ago. The values are saved, but the response is lost. I have found that the biggest improvement can be made by disabling the LED_STRIP at compile time, but i choose to keep my leds. 😃

klutvott123:
I think there is another way to improve this further. By keeping track of the states in the saving process, at timeout we can determine which steps are completed and only retry the step that failed. For example if the pagevalues have been sent to the FC and we have received the ``pageWrite`` command, but not ``eepromWrite`` when timeout occurs, we can retry only the ``eepromWrite`` command since we know that the FC has received the data.  I have implemented a rough version of this and it seems to be working well. Successful save every time.

mikeller:
Great news! And sorry, I did not mean to be overly critical, I thought we'd tried this before and hadn't had success.

Also, good idea to checkpoint the steps in the save process.

klutvott123:
Don't worry about it. I don't mind you guys being critical. In fact you should be!
I'll provide some code for review if i can get this working the way i want

PR #191, opened 2018-12-18
Discard duplicate input data for smartport

klutvott123:
Discards duplicate frames that might be in the lua input buffer. This makes the script insensitive to duplicate frames caused by RX retransmits. Testing on my setup shows a 70-80% decrease in page load retries caused by sequence number mismatch.


mikeller:
Good work identifying and addressing this!

klutvott123:
Thanks! 👍

Issue #165, opened 2018-08-29
Minor textual improvement


0crap:
Is it possible to change, in the Betaflight / Rates (3/3) page, the text Anti-Grav Thr to Anti-Grav Thr**e** ?
Thr is often referred to as Throttle.
Adding only an e makes it much more clear it's about Threshold.

Just a small suggestion. Thx!

0crap:
![20180829_191213](https://user-images.githubusercontent.com/31951195/44804420-90bad000-abc1-11e8-9da1-0afc2e4b0bac.jpg)

klutvott123:
Good suggestion. I would have changed it to "thld" which is commonly used in aviation according to https://www.allacronyms.com/threshold/abbreviated

0crap:
No need to play it by the aviation book. ;-)

Coming from rc helicopters, in my mind that is Throttle Hold.
Oh well.....

mikeller:
I like this one: https://www.allacronyms.com/THRSH/Threshold

Nine out of ten pilots will read this as 'trash'...

0crap:
trash..., yes indeed. :-)

**Thrsh** is perfect, lets give that a go.

ijustwannafly:
rates3.lua

{ t = "Anti-Grav Thrsh",  x = 15, y = 30, to = SMLSIZE },

You can add the edit now or grab the latest build from the jenkins website.

0crap:
Thx! Tested and works great.
Can you please make a Pull request so it will end up in the official code?

Much appreciated.

0crap:
No stale bot please.
Still hope this can be done.

0crap:
Merged, so can be closed.
Thx all. 👍

PR #194, opened 2019-01-03
Fix for Minor textual improvement #165

0crap:
My first pull request, so bear with me. :-)

Tested on my Taranis X7 and looks fine.

### DIFF
--- a/src/SCRIPTS/BF/X7/x7pre.lua
+++ b/src/SCRIPTS/BF/X7/x7pre.lua
@@ -1,5 +1,6 @@
-PageFiles = 
+PageFiles =
 {
+    "vtx.lua",
     "pids1.lua",
     "pids2.lua",
     "rates1.lua",
@@ -8,9 +9,6 @@ PageFiles =
     "filters1.lua",
     "filters2.lua",
     "pwm.lua",
-    "vtx.lua",
-    "rescue.lua",
-    "gpspids.lua",
 }
 
 MenuBox = { x=15, y=12, w=100, x_offset=36, h_line=8, h_offset=3 }
--- a/src/SCRIPTS/BF/X7/rates3.lua
+++ b/src/SCRIPTS/BF/X7/rates3.lua
@@ -9,7 +9,7 @@ return {
    outputBytes    = 23,
    text = {
       { t = "Anti-Grav Gain", x = 15, y = 17, to = SMLSIZE },
-      { t = "Anti-Grav Thr",  x = 15, y = 30, to = SMLSIZE },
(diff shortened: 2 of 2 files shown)


src/SCRIPTS/BF/X7/x7pre.lua
```diff
@@ -1,5 +1,6 @@
-PageFiles = 
+PageFiles =
 {
+    "vtx.lua",
```
mikeller: This ordering change is entirely subjective, and will probably be disliked by just as many users as it will be liked by, plus it will be disliked by all users that do not like a change from what they are used to, so I don't think this should be done without a good reason.

src/SCRIPTS/BF/X7/x7pre.lua
```diff
@@ -8,9 +9,6 @@ PageFiles =
     "filters1.lua",
     "filters2.lua",
     "pwm.lua",
-    "vtx.lua",
-    "rescue.lua",
```
mikeller: These should not be removed.

src/SCRIPTS/BF/X7/x7pre.lua
0crap: OMG did I do this?
I only wanted to change these on my own github 0crap account and not at all push into the official master. Only do the minor textual improvement back into the master. Can you tell me what I did wrong?

src/SCRIPTS/BF/X7/x7pre.lua
0crap: OMG did I do this?
I only wanted to change these on my own github 0crap account and not at all push into the official master. Only do the minor textual improvement back into the master. Can you tell me what I did wrong?

0crap:
I'm closing this pull request, I have no idea what went wrong.
Going to try again, sorry for the trouble.

mikeller:
@0crap: Don't close this (you can always reopen it). Git is quite good in terms of the tools available to to fix a branch (and any associated pull request).

https://github.com/betaflight/betaflight-tx-lua-scripts/pull/194/commits

Looks like you've accidentally pushed commit 'Changed order and removed GPS stuff.' into the pull request branch. `git rebase -i master`, then follow the instructions to drop the offending commit, then `git push --force-with-lease` should get you out of there.

0crap:
@mikeller Thx for your advice!

I'm still struggling to grab the GitHub concept. (For now with GitHub Desktop and less with CLI commands.)
I will replay this with a Github test account to see how above works.
I don't feel comfortable enough for now against the official BF master, so I already created a new PR.
That was my easy way out for this one. :-)

Issue #199, opened 2019-03-12
requiredVersion does not match actual API version


hsuq510:
in the gpspids.lua the string requiredVersion = 1.041 does not match the API version, therefore the gpspids is not displayed.

hsuq510:
it seems apiVersion is always 0

hsuq510:
as background is run before ui, we suggested to fix and run the check inside the main task like:


function run_ui(event)

   if not dataInitialised then
         if data_init == nil then
              data_init = assert(loadScript(SCRIPT_HOME .. "/data_init.lua"))()
         end
     dataInitialised = data_init.init();
     if dataInitialised then
           data_init = nil
           collectgarbage()
     end
   end

mikeller:
@hsuq510: Have you tested this? This will result in the UI not showing and the script locking up whenever there is no connection, which I suspect most users will find problematic.

laz2wiringpi:
i have no gps screen with build 27 and the rel  has the screen but does not work with 4.0.
IMO
 Should it even be enabled ?
 is it  coded for the changes in 4.0 ? there were many ... i might be wrong though.

laz2wiringpi:
Fyi 
revert  the code for requiredVersion for a test . 
 gps screens appear. the gps values  seem to work and the values seem to align . 
I used x7  -  requiredVersion  works on other transmitters ?

hsuq510:
> @hsuq510: Have you tested this? This will result in the UI not showing and the script locking up whenever there is no connection, which I suspect most users will find problematic.

Yes I have tested with no connection and the UI is showing , no problems in my tests

laz2wiringpi:
@hsuq510 
can you make a pull request and or a ref to your test branch ?

klutvott123:
The code works for me as it is now. BF 4.0.0 RC2, QX7 with latest lua master. The only way I can reproduce this issue is to turn on the transmitter, navigate to the UI and then power on the FC. In that case ``run_bg`` hasn't received the api version yet and the gps pages will not be visible. Since ``run_bg`` isn't allowed to run while the UI is visible, the api version won't be received until you navigate out of the UI, let ``run_bg`` do it's thing and navigate back into the UI. Then the gps pages should be there. 

I can imagine it's a little annoying/confusing. Another way to deal with it could be to allow ``run_bg`` to run until the api version is received even if the UI is visible.

hsuq510:
[...]

PR #200, opened 2019-03-17
Update ui.lua

hsuq510:
In order to fix issue 199 where api version is not initialized correctly the pull request is tested.

https://github.com/betaflight/betaflight-tx-lua-scripts/issues/199

### DIFF
--- a/src/SCRIPTS/BF/ui.lua
+++ b/src/SCRIPTS/BF/ui.lua
@@ -60,6 +60,7 @@ end
 
 local function invalidatePages()
     Page = nil
+    dataInitialised = false
     currentState = pageStatus.display
     saveTS = 0
 end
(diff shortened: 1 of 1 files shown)


src/SCRIPTS/BF/ui.lua
```diff
@@ -60,6 +60,7 @@ end
 
 local function invalidatePages()
     Page = nil
+    dataInitialised = false
```
mikeller: This shouldn't be invalidated - no point in re-reading data over and over again, especially as bandwidth is already scarce.

src/SCRIPTS/BF/ui.lua
```diff
@@ -266,6 +267,19 @@ end
 
 function run_ui(event)
     local now = getTime()
+    
+    if not dataInitialised then
+        if data_init == nil then
+          data_init = assert(loadScript(SCRIPT_HOME .. "/data_init.lua"))()
+        end
+        dataInitialised = data_init.init();
```
mikeller: Why are you running this only once? If you want to run this here, it should be run until the initialisation has finished, otherwise this will just be done after an arbitrary number of user interactions.

mikeller:
@hsuq510: I think this has been addressed in #203.

hsuq510:
@mikeller still not working with latest ui.lua data is not initialized and apiversion remain 0. At leastr with X9D is not working.

PR #202, opened 2019-03-28
Add PID Advanced features + implement scrolling

klutvott123:
Adds new features and implements scrolling in order to fit them all on one page.
The features are added to the anti gravity page. Renamed the page to ``PID Advanced`` as it's a better description of what the page contains. Everything is in the same order as in the configurator.

Features added:
- Acro Trainer Angle Limit
- Throttle Boost
- Absolute Control
- I Term Rotation
- I Term Relax
- Integrated Yaw

Did not add Smart Feed Forward because it's scheduled for removal in 4.1.
I'm also thinking about doing a filter overhaul. With scrolling it could be laid out similar to the configurator. Could probably fit all the filters with all options in one page. 

Scrolling is implemented by defining an area in the y direction. This area is defined per page by adding ``yMinLimit`` and ``yMaxLimit`` to the page file. If not defined, the limits are 0 to LCD_H.  If the current field has a y coordinate outside of these boundaries, an offset will be calculated and everything is shifted up/down so that it stays within the defined limits. It's simple, but works very well. 

Tested on my QX7. Tested in simulator for Horus X12S and Taranis X9D+.

I tried to make a video but my phone wouldn't let me. Here's some simulator screenshots instead.
It's pretty much identical for X7 and X9. Only difference is that X9 has the full text for some of the longer options.

screenshot_x7_19-03-28_19-04-08
screenshot_x7_19-03-28_19-04-22
screenshot_x7_19-03-28_19-04-26
screenshot_x12s_19-03-28_18-58-48
screenshot_x12s_19-03-28_18-58-56


klutvott123:
Thanks! 👍

McGiverGim:
Good job! If you can do the same for the filters, go for it, thanks!

klutvott123:
@McGiverGim Thank you! I will see what I can do with the filters. 🙂

PR #203, opened 2019-03-31
Fixed error if API version not received.

mikeller:


mikeller:
@hydra: I found a simple way to fix the infinite loop issue - now the behaviour is that if there are no pages with a suitable version found, the script cycles through all of the unsuitable pages as if there was no check. I don't think there is a way for the script to self-terminate - the ui routine will just be called over and over again.

codecae:
Still looking at this... since `run_ui` is being used by the telemetry script as the `run` script, [returning non-zero will cause the telem script to exit gracefully](https://opentx.gitbooks.io/opentx-2-2-lua-reference-guide/content/run_function_syntax.html#run-function-for-telemetry-scripts).

A little confused with the `local nextPage = nil` before the loop, because nothing in the loop seems to update it and it is always being compared for inequality to `currentPage`...

Will keep digging..

src/SCRIPTS/BF/ui.lua
```diff
@@ -354,9 +354,14 @@ function run_ui(event)
             incValue(-1)
         end
     end
+    local nextPage = nil
     while Page == nil do
         Page = assert(loadScript(radio.templateHome .. PageFiles[currentPage]))()
-        if Page.requiredVersion and Page.requiredVersion > apiVersion then
+        if Page.requiredVersion and apiVersion and Page.requiredVersion > apiVersion and currentPage ~= nextPage then
```
codecae: `apiVersion` will always evaluate to true of it is not nil.

> \> x = 0
> \> if x then print(1) end
1

Given that `apiVersion` is initialized as 0 and is never nil, the boolean check for this variable will always be true
https://github.com/betaflight/betaflight-tx-lua-scripts/blob/62d7d4af61811e4bb752d401c54cc96b2bea3fdc/src/SCRIPTS/TELEMETRY/bf.lua#L3

Perhaps nesting `currentPage ~= nextPage` within this if statement would permit a way to gracefully exit when they are equivalent.

mikeller: I have added exiting if no valid pages are found.

mikeller:
Anybody in for a review?

hsuq510:
didnt work on X9D, apiversion still not received.

PR #207, opened 2019-04-04
Updated ui.lua to better handle incompatible API versions

codecae:
Hoping this will help with API mismatches and prevent infinite loops of `run_ui`.

### DIFF
--- a/src/SCRIPTS/BF/ui.lua
+++ b/src/SCRIPTS/BF/ui.lua
@@ -18,6 +18,7 @@ local currentState = pageStatus.display
 local requestTimeout = 80 -- 800ms request timeout
 local currentPage = 1
 local currentLine = 1
+local availablePages = #(PageFiles)
 local saveTS = 0
 local saveTimeout = 0
 local saveRetries = 0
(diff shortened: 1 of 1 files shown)


codecae:
For some reason my amend didn't work.. trying again.

src/SCRIPTS/BF/ui.lua
```diff
@@ -355,10 +356,23 @@ function run_ui(event)
         end
     end
     while Page == nil do
-        Page = assert(loadScript(radio.templateHome .. PageFiles[currentPage]))()
-        if Page.requiredVersion and Page.requiredVersion > apiVersion then
+        if PageFiles[currentPage] then 
+            Page = assert(loadScript(radio.templateHome .. PageFiles[currentPage]))()
+            if Page.requiredVersion and apiVersion > 0 then
+                if Page.requiredVersion > apiVersion then
+                    PageFiles[currentPage] = nil
```
mikeller: This will cause problems if users are using multiple craft with different firmware versions with the same TX side model, and do not power cycle their TX in between.

codecae:
Someone would be required to power cycle or change models anyway if they want to obtain the API version, too.  The init function only runs once at the time of initial invocation, so there's a possibility that the API version retained in memory might not match the new FC.  If they do match, though, the same set of pages would still be compatible.

src/SCRIPTS/BF/ui.lua
codecae: Someone would be required to power cycle or change models anyway if they want to obtain the API version, too. The init function only runs once at the time of initial invocation, so there's a possibility that the API version retained in memory might not match the new FC. If they do match, though, the same set of pages would still be compatible.

mikeller: The init function runs every time a connection is established. As a consequence, the API version will be queried again after models have been changed.

codecae: I got the init functions confused I think... apologies.

codecae: Maybe re-seed the pages here?
https://github.com/betaflight/betaflight-tx-lua-scripts/blob/62d7d4af61811e4bb752d401c54cc96b2bea3fdc/src/SCRIPTS/BF/background.lua#L42-L46

mikeller: Sounds like a lot of complexity for not much gain, compared to what I have proposed in #203.

But adding the graceful exit makes sense.

mikeller:
@codecae: I think this has been addressed in #203.

PR #211, opened 2019-04-14
Prevent scrollover

klutvott123:
Prevents going back to the top of the page when scrolling past the end and vice versa. Does the same for the menu.
Also enables repeat press for quickly getting to the top/bottom for X9. 

Personally I find it very annoying to end up at the start after doing a quick spin on the rotary encoder to get to the bottom of the page. With this change it will just stop when you reach the first/last field.
The "downside" to this is that it's not consistent with how it works in the opentx pages. 

This is just a suggestion. If people prefer the way it currently is, I'm fine with that. 🙂

### DIFF
--- a/src/SCRIPTS/BF/ui.lua
+++ b/src/SCRIPTS/BF/ui.lua
@@ -331,9 +331,9 @@ function run_ui(event)
             incPage(-1)
         elseif event == userEvent.release.menu or event == userEvent.press.pageDown then
             incPage(1)
-        elseif event == userEvent.release.plus or event == userEvent.dial.left then
+        elseif event == userEvent.release.plus or event == userEvent.repeatPress.plus or event == userEvent.dial.left then
             incLine(-1)
-        elseif event == userEvent.release.minus or event == userEvent.dial.right then
+        elseif event == userEvent.release.minus or event == userEvent.repeatPress.minus or event == userEvent.dial.right then
             incLine(1)
         elseif event == userEvent.release.enter then
             local field = Page.fields[currentLine]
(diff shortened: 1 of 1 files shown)


lshems:
Could you add one field to every page: next page?

this would make the navigation work with only enter/exit and plus/minus??

mikeller:
@klutvott123: Not sure - I agree that wrap around is annoying when the scroll wheel is used. On the other hand, on units without a scroll wheel like the X9, wrap around is a convenient shortcut to get to the beginning of the list from near the bottom. I guess you can't have both, and be consistent.

klutvott123:
@mikeller Those were the words I've been searching for. "wrap around". "scrollover" lol.
I have been going back and forth trying to decide if I should propose these changes or not for the reason you mentioned. By enabling repeat press for radios without rotary encoders I think it will be fine. Holding plus/minus takes you to the end/start in a hurry. I have been using this for a while on my QX7 and in the simulator for X9 and I just think it's more intuitive this way. It's also nice to get a clear indication that the end of the page has been reached. If it turns out that people don't like it, we could always change it back to the way it was.

@lshems Do you mean so that you can navigate between the fields in any direction? left/right/up/down? The code as it is now isn't written with being able to add random fields with random functionality in mind. It would require some work to make that happen. Also the exit button is already used for something else.

mikeller:
@klutvott123: I think the repeat on hold is a good addition for wheel-less TX, but it does not make sense to support it and wrap-around at the same time, so we should go for this.

klutvott123:
@mikeller 👍

PR #213, opened 2019-04-21
Make it Work With Crossfire

Kai-ber:
Will it be possible to make the script work with crossfire and IRC tramp?


mikeller:
@Kai-ber: I think what you want to do is open an issue [here](https://github.com/betaflight/betaflight-tx-lua-scripts/issues), and not open a pull request (with no meaningful content).

Please close this.

Also, the script should work with Crossfire.

PR #229, opened 2019-06-22
Support of FlySky NV 14 added

qba667:
Implementation for NV14 added.
Please let me know if my change of events is correct - in my opinion plus is corresponding to right cursor movement. For navigation in NV14 currently we are using trim joysticks in following configuration:
Left:
-Up- Menu
-Down - Telemetry
-Left - PgDwn
-Right - PgUp
-Click - Exit
Right:
-Up - Plus
-Down -Minus
-Left - Rotary left 
-Right - Rotary right
-Click - Enter

In next version maybe I will be able to implement touch events.

### DIFF
--- a/src/SCRIPTS/BF/ui.lua
+++ b/src/SCRIPTS/BF/ui.lua
@@ -313,9 +313,9 @@ function run_ui(event)
     elseif currentState == pageStatus.displayMenu then
         if event == userEvent.release.exit then
             currentState = pageStatus.display
-        elseif event == userEvent.release.plus or event == userEvent.dial.left then
+        elseif event == userEvent.release.plus or event == userEvent.dial.right then
             incMenu(-1)
-        elseif event == userEvent.release.minus or event == userEvent.dial.right then
+        elseif event == userEvent.release.minus or event == userEvent.dial.left then
             incMenu(1)
         elseif event == userEvent.release.enter then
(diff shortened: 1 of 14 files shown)


src/SCRIPTS/BF/ui.lua
```diff
@@ -313,9 +313,9 @@ function run_ui(event)
     elseif currentState == pageStatus.displayMenu then
         if event == userEvent.release.exit then
             currentState = pageStatus.display
-        elseif event == userEvent.release.plus or event == userEvent.dial.left then
+        elseif event == userEvent.release.plus or event == userEvent.dial.right then
```
mikeller: These changes will cause the rotary wheel to work backwards for other TX with a rotary wheel. If the NV14's rotary wheel is mounted backwards then this has to be worked around in a different way (and should probably be fixed in OpenTX, and not in these lua scripts).

qba667: Changes reverted as requested.

qba667:
> 
> 
> Looks good now.
> 
> Now I only need an engineering sample of the NV 14 to test this. 😝
@mikeller We have to ask FlySky - NV14 is incredible controller - if not crappy LCD, few hardware issues and incomplete firmware still in progress would be perfect one.

PR #239, opened 2019-08-28
Fix cam angle max limit to match Betaflight value range

fredericmorin:
Ref: https://github.com/betaflight/betaflight/blob/master/src%2Fmain%2Fcli%2Fsettings.c/#L699


mikeller:
@raphaelcoeffic: Can I ask you to please respect the 24 hour stand down period for mergeing pull requests in Betaflight? This is to give the contributors that are scattered around the globe time to review pull requests at a time that is reasonable in their time zone.

raphaelcoeffic:
@mikeller sorry! This is probably something I missed during my longer absence. Won't happen again.

Issue #245, opened 2019-09-27
opentx 2.3 number jumping back to another value


robschii:
for example you want to switch vtx channel.......  number is jumping to the value before    or you hold the plus/minus button for a long time

hsuq510:
sames here, it seems the EVT_MINUS_BREAK is not properly handled. Any value changed with the minus button on a X9D+ (not 2019) has the effect when release the minus button the value returns the previous without any effect.

step to reproduce:
select a value, change it with minus , release minus button , see value change back to what is was at first

robschii:
nothing happend till now

RipperDrone:
https://github.com/opentx/opentx/issues/6852#issuecomment-538082073

RipperDrone:
searched in ui.lua (not sure this is the right spot...) - looks ok so far doesn't it?
![image](https://user-images.githubusercontent.com/12601105/66157632-7a810e00-e624-11e9-9ec5-435f22a81c30.png)

RipperDrone:
incValue() - is this part good?

![image](https://user-images.githubusercontent.com/12601105/66157748-b6b46e80-e624-11e9-9b73-ecadea22c929.png)

hsuq510:
betaflight event.lua has the following:

    dial = {
        left = EVT_ROT_LEFT or EVT_UP_BREAK,
        right = EVT_ROT_RIGHT or EVT_DOWN_BREAK
    }

removing the OR condition make script to work properly. This is not an opentx issue.

RipperDrone:
maybe the Horus or Taranis 2019 wheel commands are mixed up with the old button commands the wrong way...?

hsuq510:
can you please test if this in the events.lua fix the issue ?

dial = {
    left = EVT_ROT_LEFT ,
    right = EVT_ROT_RIGHT
}

RipperDrone:
No, it doesn't :-(

mikeller:
This code was added in #135 by @raphaelcoeffic. Maybe he has got an idea what needs to be done here?
As it is now with these scripts, we are supporting a number of different hardware types with different switches / toggles / wheels, so we need to thread carefully and do thorough testing to make sure we don't just shift the breakage to different hardware.

RipperDrone:
Yes, it does (I first kept the wrong half of the expression, thinking the ROT_ identifiers were supposed to fit the Taranis 2019 with ROTation wheel... :-))

[...]

PR #250, opened 2019-10-06
Virtual events (WiP)

raphaelcoeffic:
~~This is a first PoC for the virtual events. It has been only tested with T16 on the simulator.
The event checks in the UI can probably be simplified once we have something working.~~

This PR introduces support for virtual events (added to OpenTX 2.3) and implements automatic selection of the target platform (aka radio) based on the screen resolution, this supporting automatically all the radios supported in OpenTX, as long as the screen format is one of those already supported.

Let's make the tests properly. Here is a list of targets that need testing (non-exhaustive):
- [x] X9D+
- [ ] X9D+ 2019 (D16 only)
- [ ] X-lite, X-lite-S & X-lite-Pro (D16 only)
- [x] X7
- [ ] X10(S), X10(S) Express (D16 only) & T16 (using OpenTX 2.3 from current `2.3` branch)
- [ ] T12

Please make sure you have OpenTX 2.3.1 installed and unzip this file at the root of your SD card and report how it works:
[betaflight-tx-lua-scripts_1.3.0_virtual_events.zip](https://github.com/betaflight/betaflight-tx-lua-scripts/files/3720433/betaflight-tx-lua-scripts_1.3.0_virtual_events.zip)

Depending on your target, you should test executing the scripts directly from the SD manager and/or as a telemetry script. Both should work properly.


raphaelcoeffic:
This needs to run on latest opentx, which is 2.3.1 for now.

We will need tests on basically all the targets. There is a test script that @3djc wanted to update to test the virtual events (in case it does not work as expected) and potential issues need to be reported to opentx for fixing the virtual events if necessary.

Some targets with the exact same keys and hardware don’t need to be tested separately (ex: T16 / X10, all the X-lites share the same navigation).

mikeller:
@raphaelcoeffic: We are close to releasing Betaflight 4.1. Normally we do a release of the lua scripts to go with a firmware release. Do you think this is stable enough to be released this or next week? If not we should look into fixing #245 independent of this, and then releasing the scripts for now, and adding this pull request after the release.

raphaelcoeffic:
Definitely not ready yet, lot’s of tests required on many targets I don’t own.

Fixing the other issue might be just as challenging if we are to break other targets... I really don’t know what can be done best.

mikeller:
@raphaelcoeffic: I guess in this case is to delay releasing a new version of the lua scripts until we've got the virtual events at a point where we can release them. We can crowdsource the testing to the users in here, and we can add an announcement about this to the release notes for Betaflight 4.1.
So to my understanding the problem that exists with the current scripts was introduced in 2.3, so users can avoid the problem by staying with 2.2 for now?

Matze-Jung:
That's pretty interesting! I took another approach a few months ago and mapped the inputs by radio-name identifier. But from my testing, it takes some more memory.
If virtual events work well, some things will become simpler, others will no longer be in the hands of LUA script devs.

mikeller:
Since this change uses functionality that is brand new in OpenTX this will need to be tested thoroughly on as many supported hardware types as possible.
So if you are a Betaflight TX lua script user, can you please use these steps to test and provide feedback (do this for as many different TX models as you can):

- Upgrade OpenTX on your TX to 2.3.1;
- Download this build and install it on the TX by copying the folders in `obj/` over your TX SD card:

[betaflight-tx-lua-scripts_1.3.0_virtual_events.zip](https://github.com/betaflight/betaflight-tx-lua-scripts/files/3704845/betaflight-tx-lua-scripts_1.3.0_virtual_events.zip)

- run the Betaflight TX lua scripts, and
- verify that all screens on your TX look like they should;
- verify that all the buttons / scroll wheels used on your TX to control lua scripts work as they should;
- if something isn't working properly, leave a comment below stating the TX model and the problem you found;
- if all is fine, check if somebody already mentioned that everything is working for your TX model, and if not, add a comment below stating the TX model and that everything worked for you.

Thank you very much for helping us to improve the scripts!

raphaelcoeffic:
@mikeller could you please update your zip file with my latest changes? I believe this should work better. The older version showed to have some flows when testing on the simulator and X9D+.

mikeller:
@raphaelcoeffic: Done, it's `betaflight-tx-lua-scripts_1.3.0_virtual_events_2.zip`

vkolotov:
An issue here with XLite Pro running v2.3.1 OTX. Joystick events are 90 degrees inverted, e.g.:
* when you push right it actually generates "up" event (cursor goes up)
* when you push left, "down" event generated
* when you push down, nothing happens
* when you push down for 2 secs, a dialog with "save page" appears
* when you push up, screens get changed

Note: enter event works as expected

raphaelcoeffic:
@vkolotov is this related to https://github.com/opentx/opentx/pull/6916 ?

vkolotov:
> @vkolotov is this related to [opentx/opentx#6916](https://github.com/opentx/opentx/pull/6916) ?

I doubt it is related. I've copied over the new scripts and it changed direction.

The navigation in the radio controller (not in lua script) remained unchanged when I upgraded it to v2.3.1 OTX.

vkolotov:
I have just reverted the scripts (copied over the backup, the latest release) and it works as expected. So it must be something wrong in the test version (zip archive provided here).

frozenskys:
Is there a reason to still use the events.lua for mapping events if we can just use the virtual events directly in the ui.lua script?

McGiverGim:
Tested with QX7S, to me it seems that the wheel is inverted, but I'm not totally sure if it was like this before.

raphaelcoeffic:
> Is there a reason to still use the events.lua for mapping events if we can just use the virtual events directly in the ui.lua script?

In fact no, this is going away next.

@vkolotov @McGiverGim I use different events now for +/- and prev/next, so that this should now work in the proper direction (same as radio nav).

raphaelcoeffic:
Please note that the navigation is still tilted by 90° on T12 and X-lite(s, etc...). This will be fixed in OpenTx and be available in 2.3.2.

McGiverGim:
@raphaelcoeffic now the wheel works as expected.

But there is another bug (it was before too), how can I change the page? If I remember correctly, before it was with the "menu" button (the three horizontal lines), but now it does nothing. But if I go out of the Betaflight lua script, the next time I enter in I'm on a new page.

src/SCRIPTS/BF/ui.lua
```diff
@@ -327,30 +323,31 @@ function run_ui(event)
...
     -- editing value
     elseif currentState == pageStatus.editing then
-        if (event == userEvent.release.exit) or (event == userEvent.release.enter) then
+        if event == EVT_VIRTUAL_EXIT or event == EVT_VIRTUAL_ENTER then
             currentState = pageStatus.display
-        elseif event == userEvent.press.plus or event == userEvent.repeatPress.plus or event == userEvent.dial.right then
+        elseif event == EVT_VIRTUAL_INC or event == EVT_VIRTUAL_INC_REPT then
             incValue(1)
-        elseif event == userEvent.press.minus or event == userEvent.repeatPress.minus or event == userEvent.dial.left then
+        elseif event == EVT_VIRTUAL_DEC or event == EVT_VIRTUAL_DEC_REPT then
```
mikeller: This file looks so much cleaner now! :smile:

src/SCRIPTS/BF/ui.lua
```diff
@@ -302,22 +300,20 @@ function run_ui(event)
     -- process send queue
     mspProcessTxQ()
     -- navigation
-    if (event == userEvent.longPress.menu) then -- Taranis QX7 / X9
-        menuActive = 1
-        currentState = pageStatus.displayMenu
-    elseif userEvent.press.pageDown and (event == userEvent.longPress.enter) then -- Horus
+    if (event == EVT_VIRTUAL_ENTER_LONG) then
```
frozenskys: Shouldn't this be `if (event == EVT_VIRTUAL_MENU_LONG) then` ?

raphaelcoeffic: MENU_LONG cannot be used in telemetry scripts (triggers the “Reset ...” menu).

raphaelcoeffic:
> how can I change the page?

@McGiverGim I think it should be the "page" key. Have you tried it? "page long" should go back one page, and "page" should lead you to the next page.

src/SCRIPTS/BF/ui.lua
raphaelcoeffic: well, a matter of taste as it seems :-) Anyway, that saves a bit of memory by avoiding the `userEvent` structure.

McGiverGim:
> @McGiverGim I think it should be the "page" key. Have you tried it? "page long" should go back one page, and "page" should lead you to the next page.

The page key returns to the others telemetry pages, as always did I think.

McGiverGim:
Now I understand because each time I go out and in of the Betaflight script I have a different page: the page key, that rolls the telemetry pages, has been assigned to roll too the Betaflight pages. In the X7 this has been always the menu button.

raphaelcoeffic:
> The page key returns to the others telemetry pages, as always did I think.

hmm... I obviously forgot that bit... Then I need to re-instate the menu key ;-)

McGiverGim:
> hmm... I obviously forgot that bit... Then I need to re-instate the menu key ;-)

True, executing directly the script from the folders explorer, works ok. But I have it in the telemetry page. Is more intuitive to use the page key as you did, but it seems to interfere with Open TX. Maybe it can be disabled until we press Exit? Now I know because it was assigned to the menu key 😋

raphaelcoeffic:
@McGiverGim can you please re-try with my latest change?

McGiverGim:
> @McGiverGim can you please re-try with my latest change?

Now it produces an infinite loop, rolling all the pages of the Betaflight script 😮

raphaelcoeffic:
> Now it produces an infinite loop, rolling all the pages of the Betaflight script 😮

F...! that'S what happens when I don't test it before... sorry for that!

McGiverGim:
I can't test it until tomorrow. You don't need to make the same change when long pressing the menu to move page back?

frozenskys:
@raphaelcoeffic that's working a lot better, however... ENTER_LONG triggers the “Reset ...” menu **AND** the BF popup menu on my X7 :|

src/BF/bf.lua
```diff
@@ -9,6 +9,8 @@ assert(loadScript(radio.preLoad))()
 assert(loadScript(protocol.transport))()
 assert(loadScript(SCRIPT_HOME.."/MSP/common.lua"))()
 
+isTelemScript = false
```
mikeller: Extra vowels don't cost extra. :grin: How about `isTelemetryScript`?

Asizon:
Testes in X9D+, I think it take a long time for saving pages, slowly than the previus version, can it make more fast??

raphaelcoeffic:
> I think is important to have a backward key. I understand that the problem is that the page key does not work as expected. Is there a way to tell Open TX to ignore the page key until we get out of the script?

no, it is actually reserved for changing between telemetry pages.

PR #252, opened 2019-10-09
Use new OpenTX Tools screen / Add Menu / Use Virtual Events

frozenskys:
This change moves the scripts to work with the new OpenTX 2.3.x Tools screen which standardises where to find your lua scripts and makes install much easier. It also introduces a menu meaning there is no need to page through the screens, and swaps to using the new virtual events standardising the button and dial events (same as #250 but slightly different approach.)


raphaelcoeffic:
Please note that you don't have to change all the paths, you can use the new `chdir(path)` instead. And please rebase on top of master to reduce the changes (61 files changed!!!).

mikeller:
@frozenskys: I agree with @raphaelcoeffic that it will be better to leave the locations as they are, and use a 'forwarder' in `SCRIPTS/TOOLS` to invoke the script - some users will prefer to keep it as one of their telemetry screens.

Do the tools also work on X10 / X12 that currently do not have a good way to invoke scripts?

raphaelcoeffic:
> Do the tools also work on X10 / X12 that currently do not have a good way to invoke scripts?

Yes, the tools menu is present on Horus targets.

Additionally, I have a PR in place for Widgets & events: https://github.com/opentx/opentx/pull/6688
That's the next step.

frozenskys:
Ok - happy with the forwarder approach, using a folder is just an old habit from not cluttering the root of the drive, I will merge @raphaelcoeffic 's virtual event button code as that looks more complete than mine. What do you think of the menu instead of / as well as the page forward way of navigating?

raphaelcoeffic:
I like the main menu! Can I please have both 😋 more seriously, we probably need a way to make both doable, the question is how.

frozenskys:
Well if we are using two different launch scripts (for tools page and non tools page) like the bf.lua, we could set a global in the launch script to say which one to use? or we could just allow the menu to jump you in, and then page forwards and back once you have jumped in like normal, with exit taking you back to the menu?

frozenskys:
I have split this into 2 PRs one just adding the script to the tools screen, and one building on that and adding a menu for the tools screen script. Both are based off the Virtual Events change which needs merged first.

PR #253, opened 2019-10-09
Add BF Scripts to OpenTX 2.3.x Tools screen

frozenskys:
Adds the betaflight scripts to the OpenTx Tools screen introduced in 2.3. Needs the virtual event changes merged from #250 first

This was split from #252

PR #254, opened 2019-10-09
Add menu to the betaflight script when used on OpenTx 2.3 tools screen

frozenskys:
Adds a menu to the betaflight script when run from the tools screen, but not when run as a telemetry script. Needs #253 merged first.

menu_screen


mikeller:
Oooh, flash! :100:

raphaelcoeffic:
@frozenskys rebased as well. Can you please check I did not break anything? Please note you will have to "force pull", or "reset hard" on the remote ref to have the latest version and not effect a merge.

McGiverGim:
Tested and it seems to work. To me is a little unintuitive to press exit to go the menu, I supposed this will be done with the menu key, but this key cycles the betaflight pages, like in the telemetry script. So it seems the menu is an intro page and the exis is a backward key. Is ok, only I need to use it more :)

PR #256, opened 2019-10-11
Added option Arming without fix to GPS Rescue 

kaweksl:
Sometimes you just wanna fly immediately/under roof

Requires changes in betaflight ( https://github.com/betaflight/betaflight/pull/9026 )
Someone has to adjust text position for nirvana as I don't have it and there is no simulator available. 
requiredVersion should be bumped ?


mikeller:
@kaweksl:

> requiredVersion should be bumped ?

Depends on what you want. If you bump it the page won't be shown if the flight controller reports a version older than this. So in your case it is probably safe and better to not bump it.

mikeller:
@kaweksl: This needs to be rebased.

src/SCRIPTS/BF/PAGES/rescue.lua
```diff
@@ -5,8 +5,8 @@ return {
    title             = "GPS Rescue",
    reboot            = false,
    eepromWrite       = true,
-   minBytes          = 16,
-   requiredVersion   = 1.041,
+   minBytes          = 21,
+   requiredVersion   = 1.043,
```
mikeller: Seeing that we had a lot of improvements in the lua scripts since the last release, and will probably want to do another release before Betaflight 4.2 (which will bring API version 1.43) is released, I'd rather leave this at 1.041 for now - obviously the 'arming without fix' setting won't work for users before Betaflight 4.2, but at least they will still be able to use the rest of the GPS rescue page.

kaweksl: but increased minBytes won't cause problems if script will not receive 21 bytes ?

klutvott123:
GPS rescue is starting to get a little cramped 😄 

@kaweksl May I suggest laying out GPS rescue like PID Advanced? Maybe using that as a template?

kaweksl:
By like in PID Advanced you mean all settings in a single column?

klutvott123:
@kaweksl Yes. That way we could maybe fit ``Sanity Ch.`` instead of ``Snty.`` on the X7. Adding more options later is also easier. Scrolling takes care of itself. Just a suggestion of course.

kaweksl:
Yep, that would make things more readable on X7. Just wonder if do that on other radios/resolutions

klutvott123:
It doesn't hurt to have the same layout across all radios IMO. It's done like that for some of the other pages like PID advanced and filters.

mikeller:
@kaweksl: Let me know if you want to change the layout on this pull request. Apart from that this is ready for merge now.

kaweksl:
@mikeller I wanted but i'm having an issue with scrolling, looks like there is limit for how much you can scroll. So for now i just made cosmetics changes.

Do you want me to squash PR ?

mikeller:
Na, it's fine, thanks.

PR #259, opened 2019-10-13
Telemetry script fixes after #254

frozenskys:
Fixes an issue where the telemetry script does not respond to buttons on first use, introduced by #254

### DIFF
--- a/src/SCRIPTS/BF/ui.lua
+++ b/src/SCRIPTS/BF/ui.lua
@@ -438,7 +438,7 @@ function run_ui(event)
             end
         end
     end
-    if stopDisplay then
+    if stopDisplay and (not isTelemetryScript) then
         currentState = pageStatus.mainMenu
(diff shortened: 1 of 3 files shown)


src/SCRIPTS/BF/ui.lua
```diff
@@ -353,11 +353,11 @@ function run_ui(event)
                 currentState = pageStatus.editing
             end
         elseif event == EVT_VIRTUAL_EXIT then
-            if useMenu then 
-	        stopDisplay = true
-	    else
-                return protocol.exitFunc();
-	    end
+            if isTelemetryScript then 
+                return protocol.exitFunc();
```
mikeller: Stray whitespace. :space_invader:

johnelliott:
I think I just experienced this. I am going to try this on top of master or the latest release.

PR #260, opened 2019-10-14
1.4.0 fixes

raphaelcoeffic:
This is an attempt at fixing 2 issues seen with 1.4.0:
- "BF script page changed when switch between telemetry page and script":
  - `NEXT_PAGE` / `PREV_PAGE` events are now only evaluated when the script is not executed as a telemetry script.
- "values changed back after editing":
  - this is probably caused by the fixed page reading timeout (800ms) generating multiple requests.
  - when the first reply has arrived, the user starts editing the page.
  - when the next reply arrives, the changes are wiped.
  - the fix invalidates subsequent replies.

Please use this zip file to test: 
[bf_lua-1.4.0-fixes.zip](https://github.com/betaflight/betaflight-tx-lua-scripts/files/3723615/bf_lua-1.4.0-fixes.zip)

### DIFF
--- a/src/SCRIPTS/BF/MSP/common.lua
+++ b/src/SCRIPTS/BF/MSP/common.lua
@@ -129,6 +129,7 @@ function mspPollReply()
     while true do
         ret = protocol.mspPoll()
         if type(ret) == "table" then
+            mspLastReq = 0
             return mspRxReq, ret
         else
             break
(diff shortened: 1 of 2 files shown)


0crap:
@raphaelcoeffic bf_lua-1.4.0-fixes.zip contains only empty folders.

raphaelcoeffic:
@0crap sorry about that, it's fixed now.

johnelliott:
Thanks for doing this so quickly, I just experienced the bf page flipping bug and was making a video to post.

PR #262, opened 2019-10-14
Added 'release' task to Makefile.

mikeller:


Makefile
```diff
@@ -1,2 +1,7 @@
 all:
 	bin/build.sh
+
+release: all
+	FILE_NAME="betaflight-tx-lua-scripts_$$(git describe --abbrev=0 --tags).zip"; \
+	rm -f $${FILE_NAME}; \
+	zip -r $${FILE_NAME} obj/
```
raphaelcoeffic: Shouldn't we have something like:
```
cd obj/ && zip -r ../$${FILENAME} *
```

Otherwise, the files are located in `obj/*` in the archive. It seems easier to be able to unpack directly at the root of the SD card.

mikeller: Yeah we could - I just put it the way it is because that's how the zip files for all of the previous releases were organised. But we could change this, and change the instructions.

Re unpacking at the root of the SD card, I find that when doing this with the SD card inside the TX, mounted over USB, this results in the unpacking operation taking a very long time, probably due to the operating system (linux) copying the compressed files to the SD card, and then uncompressing them there. So I am not sure if I'd want to change the instructions to ask users to do it this way.

raphaelcoeffic: Ok, got it. I never noticed it was done this way.

mikeller:
@raphaelcoeffic: Should we change this to not include a top level folder, or should we leave the format for releases as they are?

raphaelcoeffic:
> @raphaelcoeffic: Should we change this to not include a top level folder, or should we leave the format for releases as they are?

Personally, I prefer not having a top folder inside the ZIP file, so I can just unzip directly on the SD card. It saves one command, but that's it.

mikeller:
@raphaelcoeffic 

> Personally, I prefer not having a top folder inside the ZIP file, so I can just unzip directly on the SD card.

Does this actually work at a reasonable speed in Windows when the SD card is mounted through the TX running OpenTX? I tried this once in linux and it was going to take 30 minutes or so, so I aborted, ending up with an incomplete set of files on the SD card.

raphaelcoeffic:
>Does this actually work at a reasonable speed in Windows when the SD card is mounted through the TX running OpenTX? I tried this once in linux and it was going to take 30 minutes or so, so I aborted, ending up with an incomplete set of files on the SD card.

I have no clue about that, I don't own an windows computer. I usually place the SD into an SD reader to do these kind of things. But otherwise it can still be copied manually if it's too slow with the mass-storage driver.

Issue #243, opened 2019-09-16
VTX values


raso12:
BF 4.1.0 the values of the vtx are not adjustable

Raimund

robschii:
Need vtx table set in betaflight

patryk-jason:
Not exactly. I've set up table correctly. Channels can be changed via LUA, power levels not.

mikeller:
@raso12: Can you please supply information on what VTX model you are using, your CLI `diff` from Betaflight, and what version of the lua scripts you are using on what TX model / firmware?

frizzle101101:
ive got same issue on x9d+ legacy, bf4.1RC5, using atx micro 3, using script version 1.3

mikeller:
@frizzle101101: Try upgrading OpenTX to 2.2.4 (or better) - there was a bug with telemetry in lua in 2.2.3: https://www.open-tx.org/2019/07/11/opentx-2.2.4

kaweksl:
same issue. I can't change any vtx settings, rest is working fine. I used vtx table for IRC Tramp EU from BF wiki.
Does the current version suppose to work with vtxtables at all ?

X9D+ legacy with opentx 2.3.1, Betaflight / KAKUTEF4V2 (KTV2) 4.1.0, bdir dshot600

kaweksl:
After some trying i found out that in BF settings you must set Band, Channel and Power to something, if not you want be able to change anything in VTX thru lua. 
Also script expect specific table, probably that hardcoded in BF 4.0.6 ? .

raphaelcoeffic:
@mikeller what is actually the status here? I don't quite understand. Is it a bug, a feature request, something that needs more doc?

mikeller:
@raphaelcoeffic: I would see this as a feature request, but we need to figure out what we want to do. With the introduction of VTX tables the old way of having hard coded VTX frequencies / power levels in the released version of the lua scripts isn't viable any more. Getting the tables dynamically from the flight controller probably isn't either, given the memory size and connection speed constraints.
Maybe we could add a facility to the Betaflight configurator that allows the user to export their custom VTX table as a .lua file, and then drop it into the script folder on their TX?
Thoughts?

raphaelcoeffic:
Well, if the tables are available / configured in the configuration, then it makes great sense to have them generated there. I’ll try to have something working in LUA. If you could give me some pointers, I could give it a try in the configuration as well.

mikeller:
[...]

PR #264, opened 2019-10-15
VTX tables

frozenskys:
Setting to draft, as this is still a work in progress and currently untested (apart from the fact it compiles and starts without error on the sim)


frozenskys:
Trying to implement the vtx tables as discussed in #243

src/BF/VTX/vtx_defaults.lua
```diff
@@ -0,0 +1,11 @@
...
+    { 5865, 5845, 5825, 5805, 5785, 5765, 5745, 5725 }, -- Boscam A
+    { 5733, 5752, 5771, 5790, 5809, 5828, 5847, 5866 }, -- Boscam B
+    { 5705, 5685, 5665, 5645, 5885, 5905, 5925, 5945 }, -- Boscam E
+    { 5740, 5760, 5780, 5800, 5820, 5840, 5860, 5880 }, -- FatShark
+    { 5658, 5695, 5732, 5769, 5806, 5843, 5880, 5917 }, -- RaceBand
+}
+frequenciesPerBand = 8
+bandTable = { [0]="U", "A", "B", "E", "F", "R" }
+deviceTable = { [1]="6705", [3]="SA", [4]="Tramp", [255]="None" }
+pitModeTable = { [0]="OFF", "ON" }
```
mikeller: Not sure if these need to be part of the 'defaults'. But then, on the other hand, it makes sense to have them in one centralised file instead of having them included the file for each TX type - we currently do too much of this anyway.

frozenskys: @mikeller I already have a plan to refactor the scripts into unified logic, and just load the "layout" for the screens from files based on the screen dimensions, rather than the radio type, as all else seems to be the same - which should make maintenance much easier. I also noticed that the rx.lua script is missing from the X7 folder - do you know if this is deliberate?

mikeller: Re missing rx.lua file, I suspect that this is just a consequence of the copy / paste way that multi TX support is done at the moment.

frozenskys:
@mikeller about the powerlevels - what do we need to send over msp back to the FC as I think with the vtx tables this has changed?

1. 0,1,2,3 (power index?)
2. 14, 20, 30 (power dBm ?)
3. 0,25,100 (power mW)
4. something else?

mikeller:
@frozenskys: The power index - the flight controller then uses the VTX table to translate this into the value to send to the VTX.

PR #266, opened 2019-10-16
Refactor to common page logic

frozenskys:
Now that OpenTx has aligned around screen size rather than device, it seems to make sense to refactor the BF lua script code in the same way.

I have tried to extract the page logic into device agnostic common pages, and the actual screen layout in to resolution dependent _"Layouts."_ As part of this I have also moved some settings up to the _radio_ level from the pages where it seemed to make sense.

Opening as a draft for comments...


frozenskys:
Gone as far as I can without feedback, ideas for further refactoring, etc.

PR #268, opened 2019-10-21
Fixes after refactoring

frozenskys:
Fixes a couple of issues after the refactor to common pages...


frozenskys:
@mikeller been thinking about the api required version (as I think we will need to change this for vtx_tables.) Does it still make sense to do it at a page level - or would it be best to pull it up to a script level and give an error if the version is mismatched?

mikeller:
@frozenskys: Not sure. Doing version checking at page level is convenient, as it allows the scripts to support (limited) backwards compatibility. If we do it at the script level then bugfixes and non-MSP-related improvements (like the new top level menu) will not become available to users of older firmware unless we backport them and do maintenance releases for older script versions.
But if version checking at page level gets too complicated / runs into memory constraints, then doing it at script level will be a way out.

PR #269, opened 2019-10-22
Fix GPS rescue layout X7

klutvott123:
After y limits was moved to radio level they are now in effect for all pages. Trying to write outside the min limit causes the text to not show up. Just moved everything to stay within the limits.


frozenskys:
@klutvott123 it works without this change on both the X7 sim and my X7 running OpenTx 2.3.1 - do you have a picture of the issue. (The y-limits should just cause screen scrolling rather than no text...)

klutvott123:
@frozenskys That's strange. I have also tested on my QX7 and in sim with 2.3.1. 
The limits were intended to work like this in order to avoid writing over the title bar.

Here's what it looks like
screenshot_x7_19-10-22_21-41-38

Here's what it should look like
screenshot_x7_19-10-22_21-42-46

Which one of these do you see?

frozenskys:
See I had made a mistake! I thought it was the bottom screen limit that was the issue - looks like your fix is good :) Sorry to doubt you, I'll check the other screens to make sure I haven't missed anything...

frozenskys:
@klutvott123 Good spot! :) - There's several other screens that suffer from this issue - did you want to take a look at them or I'm happy to do an additional PR with the other screens in...

klutvott123:
@frozenskys Happens all the time 😄 Feel free to make a PR with more fixes if you want!
I couldn't find anything other than the GPS rescue page for X7. 

I just had a quick look at the X9 pages. At least PID 2/2, rates and GPS rescue needs some work. I would suggest changing the y limits for 212x64 to be the same as for 128x64. 11 to 52. That should take care of PID 2/2 and rates.

frozenskys:
Thanks :) - Found PIDS2 / RATES / VTX on 128x64 needed some small fixes in addition to the GPS Rescure you found, also RX could do with being shuffled up the screen a little. I'm assuming the same numbers will work for 212x64 as the screen height is the same.

kaweksl:
I have fixed that in PR https://github.com/betaflight/betaflight-tx-lua-scripts/pull/256
but I'm not sure why I have started from y = 14

klutvott123:
@kaweksl  Ahh I see! Then I might close this. I usually start at y = 12 and x = 2 with a linespacing of 8 pixels because it makes good use of the limited screen space while still leaving some empty space around. Can fit 6 lines nicely without touching the titlebar or having letters cut at the bottom of the screen.

kaweksl:
Starting from y = 12 is better than 14 because Ground Speed at bottom is visible without scrolling, +1 for visible lines. I will change that in https://github.com/betaflight/betaflight-tx-lua-scripts/pull/256

klutvott123:
Fixed in #256. Closing

PR #270, opened 2019-10-22
Layout fixes

frozenskys:
Thanks to @klutvott123 for pointing out some issues with the layouts for 128x64 (QX7) and 212x64 (X9D) after the refactoring was done. This PR will need to be rebased once #268 and #269 have been merged.


klutvott123:
@frozenskys Looks like there's more that needs fixing. I'm having a look at 480x272 in the sim now. 

We need to change the text size for both high resolution screens. ``MIDSIZE`` is huge and breaks the layout for most of the pages. I think we should use the default ``textSize = 0`` as that fixes it. Default size looks better too I think. 
textSize = MIDSIZE
screenshot_x12s_19-10-23_12-10-20

textSize = 0
screenshot_x12s_19-10-23_12-10-47

frozenskys:
@klutvott123 I'm assuming that would want changed in 320x480 as well ?

klutvott123:
@frozenskys Yes that one too. 👍

frozenskys:
@mikeller rebased to latest master - do I need to squash?

mikeller:
@frozenskys: Yes please.

frozenskys:
@mikeller done - and learnt some git in the process :)

PR #284, opened 2019-11-11
VTX memory savings

klutvott123:
Have vtx_defaults.lua/modelname.lua return their tables to vtx.lua instead of using globals. This prevents the tables that are specific only to vtx.lua to be loaded all the time. Should save between 1 and 2k of memory when the vtx page is not selected.


klutvott123:
@frozenskys Done!

frozenskys:
@klutvott123 looks good - did you see @mikeller's comments on https://github.com/betaflight/betaflight-configurator/pull/1822 ? What does everyone think?

klutvott123:
@frozenskys I just moved ``deviceTable`` and ``pitModeTable`` to vtx.lua. We can always change it later without making more changes to the configurator.

frozenskys:
@klutvott123 looks spot on.. Thanks for your help on this...

PR #286, opened 2019-11-12
Added SonarCloud trigger file.

mikeller:
>


mikeller:
Na, doesn't work unfortunately.

PR #289, opened 2019-11-28
Backward compatibility

klutvott123:
The current method we're using for backward compatibility is very limited. This PR implements configurator style backward compatibility and builds further on common page logic. 
This makes it very easy for us to add or remove parameters and still have the script be fully compatible with older versions of betaflight. 

The exact same layout is now used for all radios. Templates are now used. Each screen resolution has it's own templates. Currently there's only a default template for each resolution which works fine for all the pages. It's possible to add a custom template for each page and resolution if needed. Just create the file and the page will use it instead of the default template. 
This limits what we can do with the layouts but it makes everything much easier. We don't have to manually edit 1 million coordinates anymore. Just add the new parameter whereever in a page and that's it. 

``PageFiles`` have been reworked. It now returns a table created based on ``apiVersion``. Since we're now doing a forced ``apiVersion`` detection, ``PageFiles`` will only contain compatible pages. This eliminates the need for multiple ``apiVersion/requiredVersion`` checks spread throughout the code. 
This also eliminates some bugs. Currently if a page is to be skipped because of ``requiredVersion`` we can end up with a broken main menu. Also if a page is skipped it's not possible to go back a page. These changes fixes all of that. 

Memory use is the same or maybe a little less than before. 

Did some changes to the PWM page functions. Creating tables everytime the page is drawn is very inefficient and can lead to huge memory spikes when changing values. Now empty tables are created when the page is loaded. The contents of the tables are now just overwritten instead of creating new ones. Reduces memory use quite a bit for this page. 

If anyone wants to try this: 
[betaflight-tx-lua-scripts.zip](https://github.com/betaflight/betaflight-tx-lua-scripts/files/3903181/betaflight-tx-lua-scripts.zip)


src/SCRIPTS/BF/PAGES/gpspids.lua
```diff
@@ -1,21 +1,59 @@
...
+local x = margin
+local y = yMinLim - lineSpacing
+local inc = { x = function(val) x = x + val return x end, y = function(val) y = y + val return y end }
+local labels = {}
+local fields = {}
+
+if apiVersion >= 1.041 then
+    x = margin
+    y = yMinLim - tableSpacing.header
+    labels[#labels + 1] = { t = "",         x = x, y = inc.y(tableSpacing.header) }
```
mikeller: While I like this change, and I think doing away with the per-screen-size files is a good move, these page definitions now contain a lot of repeated boilerplate.
I think in a next step we should look into ways to simplify or hide the boilerplate, so that the actual page definitions can be reduced to the things that are actually different for every page, without the repetitive elements that are the same.

klutvott123: I'll see what I can do

Issue #294, opened 2019-12-10
Larger Text Size for 480 x 272 Displays


MRC3742:
I have just tested the current master 1.5.0 build and see that you have changed to a menu driven main page. All seems to work properly for my needs on the Jumper T16 but I see that on the larger color screens (480 x 272) that the text scaling is now much smaller than with version 1.4.1. Would returning the scaling to earlier formats on the larger screen be an issue or are you looking to add additional functions per page? It is much easier to read on the larger screen if the scaling was returned to what it has been.

I understand that currently your main concern is to add features and still reduce memory usage for the Radios with limited allocated memory available. Such as the X9D+ and similar models.

Example Text Scaling on Master built Current Version
![PID_1 5 0](https://user-images.githubusercontent.com/26642502/70486064-8ffc3380-1abe-11ea-8afd-119909349679.jpg)

Example of Text Scaling on Version 1.4.1
![PID_1 4 1](https://user-images.githubusercontent.com/26642502/70486061-8e327000-1abe-11ea-8503-3ba1810605cd.jpg)

Thanks - Rich

klutvott123:
We have moved away from setting the text size per field/label. Now everything uses the same size which is set globally through one variable. Changing it will change the size of all text and might be a problem for the parameters presented in a list format. The text size is the default which is also used in the opentx menus.

Maybe it would help if the table elements were spaced a little further apart? 
You can test this yourself by changing row/col here: https://github.com/betaflight/betaflight-tx-lua-scripts/blob/f78dba804446ffc27abafa8c49fccebb551994aa/src/SCRIPTS/BF/TEMPLATES/480x272/default_template.lua#L6
This will change the spacing for all tables.

MRC3742:
It just seems with a screen more than twice the width resolution as the X9D+ that the small font is just not appropriate. I think that it can be easily adjusted using your current page views.

I've made a few changes to the 480x272 template and the radios.lua and it seems to fill the screen more to scale. It looks better and is definitely easier to read using the MIDSIZE font.

I've attached all eleven radio page screenshots in the .bmp format zipped for review.
The screenshots seem to show some text height cuttoff problems. This is only a problem in the screenshots and all text appears correctly on the Radio screen.

[Screenshots.zip](https://github.com/betaflight/betaflight-tx-lua-scripts/files/3948267/Screenshots.zip)

If this looks OK then I could do a pull request for the changes.
Thanks - Rich

[...]

PR #296, opened 2019-12-11
Larger font size and placement for the 480x272 screen

MRC3742:
Allows a larger font size and screen placement on the 480 x 272 screen sized radios.
All eleven screenshots are available for review as "Screenshots.zip" in issue #294 .

 
PID-01_MIDSIZE

### DIFF
--- a/src/SCRIPTS/BF/TEMPLATES/480x272/default_template.lua
+++ b/src/SCRIPTS/BF/TEMPLATES/480x272/default_template.lua
@@ -1,7 +1,7 @@
 return {
-    margin       = 5,
+    margin       = 10,
     indent       = 15,
-    lineSpacing  = 20,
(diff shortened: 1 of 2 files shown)


mikeller:
@klutvott123: What do you think of this?

klutvott123:
@mikeller While this looks fine and uses more of the available screen space, it also has less information on screen. 
I think that if this is merged it will only result in someone else being unhappy. What is the best text size and layout is very subjective, and personally I think it's better the way it currently is.

MRC3742:
You're right as it does have less information on screen than the smaller text and its definitely subject to a personal preference. Mine happens to be more towards the larger font.

@klutvott123 Thanks for pointing me to the area of code needing to be modified and if you feel it is not suited for future releases I will just modify it in my fork. 

Thanks Again - Rich

Issue #302, opened 2019-12-26
Unable to go back on pages with back function


Asizon:
I can't go back between pages, the back function is corrupted. I am using the latest master build, opentx 2.3.4, and Betaflight 4.2. The page feed button was also changed again and set to the 
`menu` key, I think it was more intuitive in the `page` key.

Asizon:
It is a #1835 discussion continue
https://github.com/betaflight/betaflight-configurator/pull/1835#issuecomment-564399611

klutvott123:
It's a result of moving to virtual events. The events are now the same for all TXs. It makes maintenance easier and reduces memory use. 
May I suggest trying the TOOLS version of the script instead? There the buttons will work just like expected. As a bonus it has the super nice main menu.

Asizon:
@klutvott123 The Tools version l looks awesome and works as expected thanks, but canI add this tools version to screen 1(acces with page key)???

klutvott123:
The tools version is made to be run as a one time script from the tools menu. The events won't work right if using it as a telemetry script.
I understand that it can be a little annoying to navigate to the tools page everytime it's needed but there's a lot of benefits to it. 
It uses less memory. 
All other scripts are halted when it's executed making it better able to coexist with other scripts.
Buttons work as expected.
When used in combination with the new rssi function script it offers the same functionality as the telemetry version while still using less memory.

Asizon:
Thanks for the clarification, closing this issue.

mikeller:
@klutvott123: Seeing this discussion, should we deprecate the telemetry version of the script, since it clearly has drawbacks? (I wasn't aware of the different memory footprint until now.)

klutvott123:
@mikeller I think we should. BF lua is a tool and belongs on the tools page IMO. It also makes more sense to just have one script that works on all radios. Should I make a PR for it?

mikeller:
@klutvott123: Yes please. We can probably add a `use_tools_menu_instead_of_this.txt` file just like we did in `BF/`.

klutvott123:
I'm on it. In this case I think it would be better to remove everything from the telemetry bf.lua and have it show the message about using tools menu instead. The .txt file makes sense for the ```BF``` folder as the script used to be launched directly from there, but no one will browse to the ```TELEMETRY``` folder. A lot of people will try using the telemetry version first as that's what everyone's used to. Displaying the message there ensures that everyone gets the message.

[...]

PR #306, opened 2020-01-25
Remove TELEMETRY script

klutvott123:
As discussed in #302 
Removes the telemetry version of the script.
bf.lua still exists but only displays a message telling the user to use the TOOLS version instead. bf.lua should be removed from the TELEMETRY folder at a later time.

Now that we have the script available as a one-time script in the tools menu, there aren't many good reasons to keep the telemetry version IMO. There are many benefits to just having the tools version.
Uses less memory.
Frees up a lot of memory that can be used for actual telemetry scripts.
All scripts are stopped when starting one-time script. This means more available memory for BF lua.
Keys work as expected.
Avoids potential conflicts with other scripts using the opentx telemetryPush/Pop functions.

### DIFF
--- a/src/SCRIPTS/BF/protocols.lua
+++ b/src/SCRIPTS/BF/protocols.lua
@@ -4,7 +4,6 @@ local supportedProtocols =
     {
         transport       = SCRIPT_HOME.."/MSP/sp.lua",
         rssi            = function() return getValue("RSSI") end,
-        exitFunc        = function() return 0 end,
         stateSensor     = "Tmp1",
         push            = sportTelemetryPush,
         maxTxBufferSize = 6,
@@ -16,7 +15,6 @@ local supportedProtocols =
     {
         transport       = SCRIPT_HOME.."/MSP/crsf.lua",
         rssi            = function() return getValue("TQly") end,
-        exitFunc        = function() return "/CROSSFIRE/crossfire.lua" end,
         stateSensor     = "1RSS",
         push            = crossfireTelemetryPush,
         maxTxBufferSize = 8,
--- a/src/SCRIPTS/BF/ui.lua
+++ b/src/SCRIPTS/BF/ui.lua
@@ -29,9 +29,7 @@ local saveTimeout = 0
 local saveRetries = 0
 local saveMaxRetries = 0
 local popupMenuActive = false
-local lastRunTS = 0
 local killEnterBreak = 0
-local stopDisplay = false
 local pageScrollY = 0
 local mainMenuScrollY = 0
 local PageFiles = nil
@@ -273,13 +271,6 @@ local function drawPopupMenu()
 end
 
 local function run_ui(event)
-    local now = getTime()
-    -- if lastRunTS old than 500ms
-    if lastRunTS + 50 < now then
-        invalidatePages()
-        uiState = uiStatus.init
-    end
-    lastRunTS = now
     if uiState == uiStatus.init then
         local yMinLim = radio.yMinLimit
         lcd.clear()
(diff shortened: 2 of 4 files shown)


mikeller:
One question that I have: Should the scripts in `SCRIPTS/FUNCTIONS` be running in the background all of the time, or do they have to be set up in OpenTX to be run?

I noticed that synchronisation of the RTC stopped working with the removal of the telemetry script.

klutvott123:
@mikeller It has to be set up as a special or global function. I just set it to run all the time 
```ON  Lua Script   rssi```. The RTC synchronisation could use some work. It only happens once and doesn't always work. Maybe we should do it periodically

mikeller:
@klutvott123: Ah I see, and now it's working. In this case it probably makes sense to separate the RTC function into a separate script, so that users who want RTC functionality but do not need RSSI to be injected periodically (taking up telemetry bandwidth) can do so.
As betaflight/betaflight#9382 shows there is interest in having RTC available in the firmware, and this is a way to do so that does not require extra hardware.

lshems:
So nobody is ever flying while using the BF tools script?

If he uses script on other places, they stop working during tool operation.

Just be informed

klutvott123:
@mikeller I agree. I suspect that most people only need the RTC functionality anyway. I'll see if I can split it up.

@lshems I hope not. I support using scripts like this while flying just as much as I support texting while driving. I mentioned in the first post that all scripts are stopped when running one-time scripts. 
I hope no one is relying on lua scripts to safely operate their aircraft.

RipperDrone:
I'm using the telemetry script while flying a lot, to display life tele values and to trigger haptic vibration alarms for battery low so I can feel it when flying LOS or older quads without OSD and with no or weak buzzers onboard. 

Will this still be possible with a one-time script? If no it means sacrificing some well working functionality for me...

mikeller:
@klutvott123: Not sure what you mean by 'trigger haptic vibration alarms for battery low state' - this was never something that the Betaflight TX lua scripts did. Are you confusing this with the normal reception of telemetry information, and the capability to set alarms based on them that is part of OpenTX?

RipperDrone:
@mikeller oh yeah, that might well be right :-). I'm doing both: Having all telemetry sensor information displayed on the Taranis screen inflight (which is by telemetry script), plus I have coded some alarms based on sensor information received.

So by what you clarified for me now, the more accurate question is: How can the telemetry sensor information still be displayed on the Taranis screen inflight when it is changed from telemetry script mechanism to one-time telemetry script instead?

mikeller:
@RipperDrone:

> Having all telemetry sensor information displayed on the Taranis screen inflight (which is by telemetry script)

This is also not something that the Betaflight TX lua scripts ever did - please ask these questions whoever is maintaining the script that you are using.

klutvott123:
It's just the Betaflight telemetry script that has been removed. All other telemetry/function scripts you might have should keep working as they always have. 
Keep in mind that all other scripts will be stopped to free up resources while a one time script runs https://opentx.gitbooks.io/opentx-2-2-lua-reference-guide/one-time_scripts.html. They will be started again when it's exited.

Logical switches and special functions will continue to work even if a one time script is running.

RipperDrone:
@mikeller seems i was wrong in my understanding of the concept. There is no other scripts I could ask anybody about. However, I was mistaken by the bf.lua script set as a telemetry page - thinking this is taking care of telemetry sensor values being displayed (which it is not. It's a core OTX functionality rather, it seems). All clear now by explanations given by @klutvott123. Thanks again!

image

image

PR #310, opened 2020-02-10
Memory saving + ui.lua size reduction

klutvott123:
A number of small savings that add up to 1-1.5KB.

lcdResolution/radio.resolution removed. Use highRes bool for high resolution LCDs instead.

SCRIPT_HOME removed. Use chdir("/SCRIPTS/BF") instead.

Removed some unused/unnecessary code/variables.

Made some changes to reduce the size of ui.lua. Removed unnecessary paranthesis, newlines, indents. Also removed comments where they aren't really needed.  Reducing the size of a file helps with on device compilation. ui.lua size reduced by approximately 1.5KB.

### DIFF
--- a/src/SCRIPTS/FUNCTIONS/bfbkgd.lua
+++ b/src/SCRIPTS/FUNCTIONS/bfbkgd.lua
@@ -1,8 +1,8 @@
-SCRIPT_HOME = "/SCRIPTS/BF"
+chdir("/SCRIPTS/BF")
 apiVersion = 0
-protocol = assert(loadScript(SCRIPT_HOME.."/protocols.lua"))()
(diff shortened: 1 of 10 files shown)


src/SCRIPTS/FUNCTIONS/bfbkgd.lua
```diff
@@ -1,8 +1,8 @@
-SCRIPT_HOME = "/SCRIPTS/BF"
+chdir("/SCRIPTS/BF")
```
mikeller: Good idea!

klutvott123: 👍

PR #315, opened 2020-02-18
Prevent double compile

klutvott123:
Forces a recompile of scripts_compiled.lua after writing to it. This ensures that the .lua and .luac are the same and prevents a potential recompile the next time bf.lua is launched.


mikeller:
Turtles all the way...

PR #317, opened 2020-02-23
VTX Table download

klutvott123:
Downloads vtx tables if  supported and available. This is done as part of ui.lua's init state.

The vtx tables are saved as /BF/VTX/modelname.lua where "modelname" is the name of the currently selected model. Delete this file if the vtx tables need to be updated for some reason. The script will download them again.


mikeller:
One thought: What does the user do if they have changed the VTX table? Should we add a 'discard VTX table' menu item / checkbox to force-reload?

PR #318, opened 2020-02-29
Add vtx tables download to popup menu

klutvott123:
Deletes the contents of the .lua file containing vtx tables for the selected model and downloads them again.
The menu item is only available for firmwares supporting vtx tables.


klutvott123:
Just did some minor fixing. Wasn't working exactly how i wanted. Now it's good. Also a little clearer what's going on.

PR #335, opened 2020-05-03
Remove Gyro/PID loop

klutvott123:
for Betaflight 4.2 and up. With gyro native rate sampling these numbers will only make sense for gyros with 8kHz sampling rate.

### DIFF
--- a/src/SCRIPTS/BF/PAGES/pwm.lua
+++ b/src/SCRIPTS/BF/PAGES/pwm.lua
@@ -38,8 +38,10 @@ if apiVersion >= 1.031 and apiVersion <= 1.040 then
 end
 
 if apiVersion >= 1.016 then
-    fields[#fields + 1] = { t = "Gyro Update",     x = x,      y = inc.y(lineSpacing), sp = x + sp, min = 1, max = 32, vals = { 1 }, table = {}, upd = function(self) self.updatePidRateTable(self) end }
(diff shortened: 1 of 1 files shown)


src/SCRIPTS/BF/PAGES/pwm.lua
```diff
@@ -37,9 +37,12 @@ if apiVersion >= 1.031 and apiVersion <= 1.040 then
     fields[#fields + 1] = { t = "32kHz Sampling",  x = x,      y = inc.y(lineSpacing), sp = x + sp, min = 0, max = 1, vals = { 9 }, table = { [0] = "OFF", "ON" }, upd = function(self) self.updateRateTables(self) end }
 end
 
-if apiVersion >= 1.016 then
+if apiVersion >= 1.016 and apiVersion <= 1.042 then
```
mikeller, line 40: <nitpick>
Tiny little bit cleaner with:
```
if apiVersion >= 1.016
	if apiVersion <= 1.042 then
￼    	fields[#fields + 1] = { t = "Gyro Update",     x = x,      y = inc.y(lineSpacing), sp = x + sp, min = 1, max = 32, vals = { 1 }, table = {}, upd = function(self) self.updatePidRateTable(self) end }
￼    	fields[#fields + 1] = { t = "PID Loop",        x = x,      y = inc.y(lineSpacing), sp = x + sp, min = 1, max = 16, vals = { 2 }, table = {} }
	￼end
￼
	[...]
```
</nitpick>

klutvott123: Done! 👍

PR #338, opened 2020-05-09
Update readme

klutvott123:
Added instructions for background script.
Removed 4.1.0 requirement.
Merged "How to download the scripts" into "Installing" section.
some minor changes.


jforhan:
> Added instructions for background script.
> Removed 4.1.0 requirement.
> Merged "How to download the scripts" into "Installing" section.
> some minor changes.

I see the changes add the info about not including the "OBJ" directory, but...

Looking at the latest 1.4.1 release ZIP directory structure, your bf.lua/bf.luac files will never end up in /SCRIPTS/TOOLS as those files are in a directory named "BF" under the root "OBJ" directory. The other directories will merge when you paste but you end up with a "BF" folder in the root of your SD card. Either the instructions should be changed or there should be additional steps added to copy these two files to the /SCRIPTS/TOOLS directory...Unless the directory structure will be changing in the next release? 

Thanks
Jeff

klutvott123:
@jforhan These instructions are for 1.5.0. In 1.5 bf.lua will be in /SCRIPTS/TOOLS. The BF folder has been repurposed for storing vtx tables.

jforhan:
> @jforhan These instructions are for 1.5.0. In 1.5 bf.lua will be in /SCRIPTS/TOOLS. The BF folder has been repurposed for storing vtx tables.

Thanks for the info. I am guessing the existing instructions are not going to be updated as the 1.5 will be out soon enough?

Thanks
Jeff

PR #339, opened 2020-05-10
Added initial CMS framework (CRSF Only)

codecae:
Initial CMS framework and tool script.  Currently for CRSF only.

Blocked by CMS over CRSF Compression PR found [here](https://github.com/betaflight/betaflight/pull/9746)


codecae:
Not likely.  Can't have this without the other PR.  Didn't make the cut for 4.2 due to timing on my behalf.

codecae:
Also need to work in screen definitions for hi-res radios. (Horus, t16, etc)

codecae:
Added support for 480x272 screens (Jumper, Horus, etc)

src/SCRIPTS/BF/CMS/common.lua
```diff
@@ -0,0 +1,120 @@
+local _ = {
```
mikeller, line 1: What's the reason to name this `_`? I thought convention in lua was to use `_` for unused values

src/SCRIPTS/BF/radios.lua
```diff
@@ -19,6 +37,24 @@ local supportedRadios =
...
+            yOffset = 0,
+            textSize = SMLSIZE,
+            refresh = {
+                event = EVT_VIRTUAL_INC,
+                text = "Refresh: [+]",
+                top = 1,
+                left = 156,
+            },
+            exit = {
+                event = EVT_VIRTUAL_EXIT
```
mikeller, line 55: Seems to be the same everywhere - does this need to be configured per screen size, now that we have virtual events?

src/SCRIPTS/BF/CMS/common.lua
```diff
@@ -0,0 +1,120 @@
...
+    batchId = 0,
+    sequence = 0,
+    reset = function()
+        screen.buffer = {}
+        screen.data = {}
+        screen.batchId = 0
+        screen.sequence = 0
+    end,
+    draw = function()
+        if (#screen.buffer ~= nil and screen.config ~= nil and #screen.buffer > 0) then
```
klutvott123, line 51: This is confusing. ``#screen.buffer`` will always not be nil as long as ``screen.buffer`` exists. This will make the script crash if ``screen.buffer`` is nil for some reason. Did you mean ``screen.buffer ~= nil`` maybe?

src/SCRIPTS/TOOLS/bf.lua
```diff
@@ -9,7 +9,7 @@ local scriptsCompiled = assert(loadScript("COMPILE/scripts_compiled.lua"))()
 if scriptsCompiled then
     protocol = assert(loadScript("protocols.lua"))()
     radio = assert(loadScript("radios.lua"))()
-    assert(loadScript(protocol.transport))()
+    assert(loadScript(protocol.mspTransport))()
```
klutvott123, line 12: Has to be changed here too https://github.com/betaflight/betaflight-tx-lua-scripts/blob/61a860fcdd94ffb92362ae471966f3b17c91046f/src/SCRIPTS/FUNCTIONS/bfbkgd.lua#L4

src/SCRIPTS/BF/radios.lua
klutvott123: It would be great if ``cmsConfig`` could be loaded only for the CMS version since it's not needed for the MSP version.

src/SCRIPTS/BF/CMS/common.lua
codecae: This was just a convenient way of bucketing constants into a single reference object, while limiting keystrokes and improving readability.  Open to suggestions.

src/SCRIPTS/BF/CMS/common.lua
```diff
@@ -0,0 +1,120 @@
...
+        sequence = 2,
+        data = 3
+    }
+}
+
+local function cRleDecode(buf)
+    local dest = {}
+    local rpt = false
+    local c = nil
+    for i=1, #buf do
```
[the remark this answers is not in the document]
codecae: Will address

src/SCRIPTS/BF/CMS/common.lua
codecae: Looks like a mistake.  I'll take a look.

src/SCRIPTS/BF/CMS/common.lua
mikeller: `CONST`?

src/SCRIPTS/BF/radios.lua
codecae: can they be called separate versions if they're based on the same framework?

codecae: @mikeller this virtual event is kind of an assertion, as I don't have all of the hardware to confirm usability in this context.   I'd like to have this tested on multiple platforms to see if it can truly be unified.

codecae: @klutvott123 would the suggestion then be to have radiosMsp and radiosCms?  I feel like having them in one object is a little more DRY in nature.

src/SCRIPTS/TOOLS/bf.lua
codecae: addressed here:
https://github.com/betaflight/betaflight-tx-lua-scripts/blob/7285a6708da4f55759f7e38c0c52cbcd3f349abc/src/SCRIPTS/FUNCTIONS/bfbkgd.lua#L4

src/SCRIPTS/BF/radios.lua
codecae: would it be better to put the MSP related attributes in a nested `msp` table and the cms related attributes into a nested `cms` table (instead of `cmsConfig`)?

klutvott123: @codecae I'm using the word "version" loosely 😁. I was thinking maybe pass a variable to radios.lua when executing it and put cms or msp stuff into the ``radio`` table based on that.

codecae: I added an update to separate the `msp` vs `cms` attributes in the radio config.  I think this is a good compromise.  Let me know what you think.

klutvott123: @codecae Looks good to me! Thanks! 👍

codecae:
Will do 👍

PR #341, opened 2020-05-11
Change init message

klutvott123:
from "Initialising" to "Waiting for connection" to let people know that a connection to the FC is needed for it to work.
Changed telemetry version message to "Use TOOLS menu" to fit the smaller screens.

### DIFF
--- a/src/SCRIPTS/BF/ui_init.lua
+++ b/src/SCRIPTS/BF/ui_init.lua
@@ -11,7 +11,7 @@ end
 
 local function init()
     if apiVersion == 0 then
-        lcd.drawText(6, radio.yMinLimit, "Initialising")
+        lcd.drawText(6, radio.yMinLimit, "Waiting for connection")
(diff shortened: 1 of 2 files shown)


klutvott123:
Fits nicely on the QX7 👍 
screenshot_x7_20-05-11_20-14-28

0crap:
Looks great! These two small text changes make it much better for the smaller screens.

0crap:
It's disappointing not to see this in milestone 1.5
@mikeller can you explain why if I may ask?

mikeller:
@0crap: Trying to avoid getting even more delay due to feature creep - the reality is that there is always a 'very nice next feature', and waiting for it will delay any release infinitely.

0crap:
@mikeller I fully understand that general remark. But in this case it's not a new feature or something like that. Just make the huge overhaul of 1.5.0 better understandable for the public when they first see it.
Especially because there is no RC phase I would have added this before.
But again, I understand you have more on your mind besides this. Just IMHO.

PR #361, opened 2020-09-08
Craft name for vtx tables

kristjanbjarni:
Reads craft name from MSP and uses that name to store the vtx tables. If the craft name is empty it falls back on using the transmitter model name as before. This should solve the problem of using an incorrect vtx table when people have many different crafts with different vtx hardware under the same model name.

### DIFF
--- a/src/SCRIPTS/BF/ui_init.lua
+++ b/src/SCRIPTS/BF/ui_init.lua
@@ -1,13 +1,6 @@
 local apiVersionReceived = false
 local vtxTablesReceived = false
-local data_init, getVtxTables
-local vtxTables = loadScript("/BF/VTX/"..model.getInfo().name..".lua")
-
-if vtxTables and vtxTables() then
-    vtxTablesReceived = true
-    vtxTables = nil
-    collectgarbage()
-end
+local data_init, getVtxTables, getMCUId
 
 local function init()
     if apiVersion == 0 then
@@ -18,6 +11,18 @@ local function init()
         data_init = nil
         apiVersionReceived = true
         collectgarbage()
+    elseif apiVersion >= 1.042 and not mcuId then
+        lcd.drawText(6, radio.yMinLimit, "Waiting for device ID")
+        getMCUId = getMCUId or assert(loadScript("mcu_id.lua"))()
+        if getMCUId() then
+            getMCUId = nil
+            local vtxTables = loadScript("/BF/VTX/"..mcuId..".lua")
+            if vtxTables and vtxTables() then
+                vtxTablesReceived = true
+                vtxTables = nil
+            end
+            collectgarbage()
+        end
     elseif apiVersion >= 1.042 and not vtxTablesReceived then
         lcd.drawText(6, radio.yMinLimit, "Downloading VTX Tables")
         getVtxTables = getVtxTables or assert(loadScript("vtx_tables.lua"))()
@@ -29,7 +34,7 @@ local function init()
     else
         return true
     end
-    return apiVersionReceived and vtxTablesReceived
+    return apiVersionReceived and vtxTablesReceived and mcuId
 end
 
 return init
--- a/src/SCRIPTS/TOOLS/bf.lua
+++ b/src/SCRIPTS/TOOLS/bf.lua
@@ -2,6 +2,7 @@ local toolName = "TNS|Betaflight setup|TNE"
 chdir("/SCRIPTS/BF")
 
 apiVersion = 0
+mcuId = nil
 
 local run = nil
 local scriptsCompiled = assert(loadScript("COMPILE/scripts_compiled.lua"))()
--- a/src/SCRIPTS/BF/PAGES/vtx.lua
+++ b/src/SCRIPTS/BF/PAGES/vtx.lua
@@ -1,7 +1,6 @@
-local md = model.getInfo();
-local vtx_tables = loadScript("/BF/VTX/"..md.name..".lua")
-if vtx_tables then
-    vtx_tables = vtx_tables()
+local vtx_tables
+if apiVersion >= 1.042 then
+    vtx_tables = assert(loadScript("/BF/VTX/"..mcuId..".lua"))()
 else
     vtx_tables = assert(loadScript("/BF/VTX/vtx_defaults.lua"))()
 end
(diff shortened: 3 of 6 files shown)


src/SCRIPTS/BF/ui_init.lua
```diff
@@ -18,6 +22,19 @@ local function init()
...
+        craftNameReceived = getCraftName()
+        if craftNameReceived then
+            getCraftName = nil
+            vtxTables = loadScript("/BF/VTX/"..craftName..".lua")
+            if vtxTables and vtxTables() then
+               vtxTablesReceived = true
+               vtxTables = nil
+            end
+            collectgarbage()
+         end
```
klutvott123, line 37: I think downloading ```craftName``` should be conditional to ```apiVersion >= 1.042``` since it's only needed for firmwares that support vtx tables. This will save some time at init for older firmwares.

```craftNameReceived``` should be made local to this file so that it's garbage collected when init is done.

**Comment by kristjanbjarni** [comment]
Addressed all of the review comments by removing dead code and adding check for conditional download of craft name if vtx tables are not supported.

**Comment by klutvott123** [comment]
@kristjanbjarni Looks good 👍. Tested and works as expected. Could you squash please?

**Comment by klutvott123** [comment]
There is still the issue that craft name isn't guaranteed to be unique just like model name. I think the best solution would be if each FC had a unique identifier. STM devices have a unique 96 bit id and it's available through MSP(MSP_UID). This could possibly simplify things and remove a lot of confusion. 

@mikeller Any thoughts on using the mcu id for this?

**Comment by kristjanbjarni** [comment]
@klutvott123 This is my first pull request so I am not really sure how I squash correctly, seems I messed it up?
 
> There is still the issue that craft name isn't guaranteed to be unique just like model name. I think the best solution would be if each FC had a unique identifier. STM devices have a unique 96 bit id and it's available through MSP(MSP_UID). This could possibly simplify things and remove a lot of confusion.
> 
I think the only reason for craft name not being unique is if somebody is using it for their pilot name. Since there already exists "Display name" for pilot name I would consider that to be erroneous usage of craft name. But maybe mcu id would be a better idea.

**Comment by mikeller** [comment]
@klutvott123: Good idea! Using the MCU id will solve this once and for all.

**Comment by klutvott123** [comment]
All right let's do it! 

@kristjanbjarni I agree, but there will always be someone doing it wrong. By using the mcu id it should just work without the user having to keep track of all their model names and craft names. The code you have already written could easily be adapted to get the mcu id from the FC. Would you like to have a go at it?

You had two commits on your branch and wanted to squash them into one commit. 
You would do this with ```git rebase -i HEAD~2```. This brings up a text editor with a list of the two last commits prepended with ```pick```along with some instructions. Change ```pick``` to ```squash```(but ```fixup``` might be better in this case) for the second commit in the list, save and exit. Now you should have only one commit on your local branch. Now to make the remote branch the same as your local branch you can do ```git push --force-with-lease```.

I assume your local branch now has the same commit history as the remote. 3 commits and a merge. To get it back to where you were before squashing/pushing/pulling/merging do you can do ```git log``` to find the hash string of the ```Simplified ui_init``` commit. Then you can do ```git reset --hard your-commit-hash-string-here```. Then you should be back where you started and can do the process described above.

**Comment by kristjanbjarni** [comment]
@klutvott123 I fixed the commits.

Yes I can look into changing the code to use MCU id instead.

**Comment by kristjanbjarni** [comment]
OK changed the name to use instead the MSP_UID as 24 hex characters to make it short and consistant. Tested a couple of quads, seemed to work just fine.

File: src/SCRIPTS/TOOLS/bf.lua [hunk 2]
```diff
@@ -2,6 +2,7 @@ local toolName = "TNS|Betaflight setup|TNE"
 chdir("/SCRIPTS/BF")
 
 apiVersion = 0
+craftName = nil
```
klutvott123, line 5: We would have to change this to ```mcuId``` and also update all the other variables/filenames etc that contains ```craftName```.

kristjanbjarni:
Updated to use mcuId and fixed the order of the id to be the same order as mcu_id in CLI. I removed the check for apiVersion>1.042 since although you can't download the vtx tables under that version you should still be able to manually copy them to the transmitter and use them with the MCU id, right?

klutvott123:
@kristjanbjarni I think we should keep the check. Vtx tables were introduced in betaflight 4.1(api version 1.042) so I don't see why we would want to load vtx tables for firmware versions older than that. For that we have ```vtx_defaults.lua```.

kristjanbjarni:
@klutvott123 Added the check back in

src/SCRIPTS/BF/ui_init.lua
```diff
@@ -18,6 +11,22 @@ local function init()
...
+                local vtxTables = loadScript("/BF/VTX/"..mcuId..".lua")
+                if vtxTables and vtxTables() then
+                    vtxTablesReceived = true
+                    vtxTables = nil
+                end
+                collectgarbage()
+            end
+        else
+            mcuId = model.getInfo().name
+        end
```
klutvott123, line 29: Not sure if storing the model name in ```mcuId``` is the best thing to do. I would combine the conditionals like 
```elseif apiVersion >= 1.042 and not mcuId then```
and leave ```mcuId = nil```for older versions.
This would require loading vtx tables or defaults in vtx.lua based on ```apiVersion``` like this:
```
local vtx_tables
if apiVersion >= 1.042 then
    vtx_tables = assert(loadScript("/BF/VTX/"..mcuId..".lua"))()
else
    vtx_tables = assert(loadScript("/BF/VTX/vtx_defaults.lua"))()
end
```

Can we change the message to just "Waiting for device ID"? That would make it fit nicely on the 128x64 screens too.

klutvott123:
@kristjanbjarni Great! Just added some comments. We're almost there 😁

klutvott123 (approved):
👍

mikeller (approved):
Nice work, thanks @kristjanbjarni and @klutvott123, this will put this problem to an end once and for all.

klutvott123:
Edited PR title to make it clear that we're using mcu ID and not craft name. Just to avoid any confusion

limonspb:
Sorry, late to the party and got nothing usefully to say, but thanks a lot!!! This is brilliant.

PR #363, opened 2020-10-02
Allow drawing on whole display

klutvott123:
Allow drawing from 0 to LCD_H instead of using the scroll limits.
This makes the scrolling a little more "fluid" since it avoids lines of text just suddenly appearing or disappearing. It also shows the user that there's more on the page. 

Before:
screenshot_x7_20-10-02_21-10-06
After:
screenshot_x7_20-10-02_21-11-03


mikeller (approved):
Good change!

PR #365, opened 2020-10-07
Rates type support

klutvott123:
Adds support for changing rates type.

Works like in the configurator. Changes labels, scaling, min, max, and applies default values for the selected rates type.

For testing: [betaflight-tx-lua-scripts_1.5.0.zip](https://github.com/betaflight/betaflight-tx-lua-scripts/files/5340275/betaflight-tx-lua-scripts_1.5.0.zip)

### DIFF
--- a/src/SCRIPTS/BF/ui.lua
+++ b/src/SCRIPTS/BF/ui.lua
@@ -358,6 +358,9 @@ local function run_ui(event)
             end
         elseif pageState == pageStatus.editing then
             if event == EVT_VIRTUAL_EXIT or event == EVT_VIRTUAL_ENTER then
+                if Page.fields[currentField].postEdit then
+                    Page.fields[currentField].postEdit(Page)
+                end
                 pageState = pageStatus.display
             elseif event == EVT_VIRTUAL_INC or event == EVT_VIRTUAL_INC_REPT then
                 incValue(1)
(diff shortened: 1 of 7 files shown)


Asizon:
Wow! Awesome work @klutvott123 , thankyou!

klutvott123:
@Asizon Thanks! 😸

src/SCRIPTS/BF/ui.lua
```diff
@@ -358,6 +358,7 @@ local function run_ui(event)
             end
         elseif pageState == pageStatus.editing then
             if event == EVT_VIRTUAL_EXIT or event == EVT_VIRTUAL_ENTER then
+                pcall(Page.fields[currentField].postEdit, Page)
```
mikeller, line 361: Why do we need the `pcall()` here?

mikeller (approved):

mikeller:
Nicely done!

src/SCRIPTS/BF/ui.lua
klutvott123: It just prevents the script from crashing if the field doesn't have ```postEdit```. 
Could have done this too
```
if Page.fields[currentField].postEdit then
    Page.fields[currentField].postEdit(Page)
end
```

mikeller: I actually think this is a better way, as it is more specific. Otherwise any error that snuck into a `postEdit()` function could go undetected for a long time, causing debug nightmares.

klutvott123: Good point! I'll change it

mikeller (approved):

PR #375, opened 2020-12-18
Accelerometer calibration

### LINKED ISSUE #290: Add calibration
With the requirement to have a valid ACC calibration if ACC is enabled introduced in betaflight/betafligh#9031, some users will be caught out, finding themselves in the field without a calibrated ACC, CMS not available (LoS pilots), and not able to figure out the required stick commands.
betaflight/betaflight#9179 addressed some of this by adding calibration to CMS. To cover all bases we should add a similar menu to the lua scripts, or add the calibration to the main menu.

@klutvott123: Any suggestions as to where to best put this? Separate page / add to a page / menu / ...

klutvott123:
Adds a special page to calibrate the accelerometer. The page can be accessed from the menu like this:
screenshot_x7_20-12-19_11-46-19
selecting "acc cal" brings up this:
screenshot_x7_20-12-19_11-46-24
where [ENTER] is the enter key and [EXIT] is the exit key on the radio. Pressing the enter key will send ```MSP_ACC_CALIBRATION``` to the flight controller. The flight controller will calibrate the accelerometer and send the command back. The page will exit automatically when the response is received.

Fixes #290

### DIFF
--- a/src/SCRIPTS/BF/ui.lua
+++ b/src/SCRIPTS/BF/ui.lua
@@ -16,7 +16,7 @@ local pageStatus =
 local uiMsp =
 {
     reboot = 68,
-    eepromWrite = 250
+    eepromWrite = 250,
 }
 
 local uiState = uiStatus.init
@@ -72,6 +72,13 @@ local function eepromWrite()
     protocol.mspRead(uiMsp.eepromWrite)
 end
 
+local function accCal()
+    invalidatePages()
+    currentField = 1
+    Page = assert(loadScript("Pages/accelerometer.lua"))()
+    collectgarbage()
+end
+
 local function getVtxTables()
     uiState = uiStatus.init
     PageFiles = nil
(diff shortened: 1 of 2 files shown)


src/SCRIPTS/BF/ui.lua
```diff
@@ -85,6 +90,7 @@ local function createPopupMenu()
         { t = "save page", f = saveSettings },
         { t = "reload", f = invalidatePages },
         { t = "reboot", f = rebootFc },
+        { t = "acc cal", f = accCal },
```
mikeller, line 93: Not sure if we want to make this a bit more verbose, like pop up a dialog saying something like 'Make sure the craft is level and stable, then press [Enter] to continue, or [Cancel]'?
Otherwise I suspect unsuspecting newbie pilots will 'test' this function and then end up un-calibrating their craft when they 'try' this.

klutvott123: Good point! I'll see what I can do. I think the best approach would be to just make it a special page that can only be entered from the popup menu. I looked into the built in popup dialogue in opentx but it's very limited and behaves differently on different radios

klutvott123: @mikeller It's done. I updated the first post with a description of how this works with pictures

mikeller (approved):

PR #385, opened 2021-04-08
Do not work on Inav 2.6.1 OTX 2.3.11 and Radiomaster TX16S

Chefhandwerker:
Do not work on Inax 2.6.1/ OTX 2.3.11 on Radiomaster TX16s
Can you make it work again?


klutvott123:
No. The Betaflight lua scripts have never officially supported INAV and I don't recommend using it with INAV as it can result in unintentional changes being made to your config. A lot of the parameters that can be changed with the scripts are specific to Betaflight. The best solution would be for someone to make a fork and adapt the scripts to work with INAV.

PR #403, opened 2021-10-25
Remove custom page template support

klutvott123:
Removes support for custom page templates. We're not using it and we don't need it.
The idea was to be able to have custom layouts for pages/screen resolutions, but so far we have been doing just fine with the default templates.


mikeller (approved):

mikeller:
Nice cleanup!

PR #424, opened 2022-01-16
CMS fixes

klutvott123:
Various fixes for the CMS script.

- Removed double buffering for incoming data. It's not needed
- Removed some redundant tests
- Removed clearing of screen buffer so that it can be drawn multiple times
- Made some global variables local
- Fixed retry timeout reset
- Draw screen two times instead of one to avoid flickering on colour LCDs that use double buffering

 For testing
[betaflight-tx-lua-scripts_1.5.0.zip](https://github.com/betaflight/betaflight-tx-lua-scripts/files/7877901/betaflight-tx-lua-scripts_1.5.0.zip)

### DIFF
--- a/src/SCRIPTS/BF/CMS/common.lua
+++ b/src/SCRIPTS/BF/CMS/common.lua
@@ -41,31 +41,26 @@ screen = {
     data = {},
     batchId = 0,
     sequence = 0,
+    redraws = 2,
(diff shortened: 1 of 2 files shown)


haslinghuis (approved):
Tested on Taranis X9D+ ACCST

PR #419, opened 2022-01-07
Ghost MSP support

daleckystepan:
Add MSP support for Ghost

Need new Ghost FW v1.0.6.0+

BF: https://github.com/betaflight/betaflight/pull/11242
ETX: https://github.com/EdgeTX/edgetx/pull/1377

PR #429, opened 2022-01-22
Read lua buffer until empty

klutvott123:
Reads everything from the lua input buffer in one lua cycle.

The current behaviour is to read one frame from the buffer each lua cycle which is approximately every 50ms.

With this change the script will read frames from the buffer until it's empty. This can speed up the page loading significantly depending on the protocol used. For smartport/fport/crossfire it will likely not have a big effect. For smartport we get a lot of small frames at a slow interval and for crossfire the frames are large so the whole payload fits in one or two frames.
For ghost(#419) on the other hand we get small frames at a fast rate so it's possible that the buffer fills up quickly and there are actually multiple frames in the buffer so it makes sense to read them all instead of one small frame every 50ms.

### DIFF
--- a/src/SCRIPTS/BF/MSP/common.lua
+++ b/src/SCRIPTS/BF/MSP/common.lua
@@ -100,25 +100,24 @@ function mspReceivedReply(payload)
     end
     if idx > protocol.maxRxBufferSize then
         mspRemoteSeq = seq
-        return true
+        return false
     end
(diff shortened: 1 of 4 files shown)


klutvott123:
This is ready to go. It's important for ghost performance

haslinghuis (approved):

Issue #230, opened 2019-06-24
Ability to change PID profile in script


andmiller:
It would be convenient to be able to change the PID profile via the LUA script.

kafie1980:
Here I mention about a lua script (not written by me) that probably still works on X9d size transmitters: https://github.com/betaflight/betaflight-tx-lua-scripts/issues/290#issuecomment-690254544

You should be able to change the PID profile with it and other stick based commands.

It does not work for X-Lite S probably due to screen size.

Issue #47, opened 2017-10-08
changing PID/Rates profile - feature request 


fr3shn3ss:
Not sure if this is a right place to request a feature. Could someone add a feature to change PID/Rates profiles. It's really useful for racers to switch the racing PID/Rates to free style.

Issue #22, opened 2017-03-17
Great but missing some features 


Dutchy-79:
This is really great. 

Only thing is I miss the option to show profile 1 2 and 3.

And when switching to another profile with the sticks the page isn't updated, you need to change page and back to update it. 

I'd like to see it showing the selected profile and update the numbers when changing profile. 

Regards,

beeb:
Additionnally, I think it would be nice, if technically feasible, to be able to switch between rate profiles (on top of pid profiles). Cheers

mikeller:
@beeb: Just in case you are not aware: It's possible to configure rate profile switching onto a 3 position switch through the 'Adjustments' tab, and then switch in-flight.

beeb:
@mikeller absolutely! The only problem is that I have no free channel on my RX for this (all occupied by flight mode, buzzer, reverse dshot direction and arm switch). I could re-bind my RX with 16 channels but I heard it increases the latency slightly over 8 channels.

mikeller:
@beeb: There are plans to address this shortcoming in 3.3. Watch this space. ;-)

PR #432, opened 2022-02-03
Add Profiles page.

### LINKED ISSUE #430: Lua script motor output limit adjustment
### Is your feature request related to a problem? Please describe

There is no option to change motor output limit or PID profile in the Lua script.

### Describe the solution you'd like

Add an option to change motor output limit to a submenu of the lua script. 

### Describe alternatives you've considered

OSD menu from the DJI goggles don’t provide an option to adjust. A field PC is needed to change. Changes to master multiplier or throttle output won’t prevent voltage spikes due to max motor output.  

### Other information

_No response_

klutvott123:
Adds a page with pid and rate profile.
Because of how it's implemented in the firmware, both profiles can't be changed at the same time. 

Fixes #430 #230 #47 #22

### DIFF
--- a/src/SCRIPTS/BF/pages.lua
+++ b/src/SCRIPTS/BF/pages.lua
@@ -5,6 +5,10 @@ if apiVersion >= 1.036 then
 end
 
 if apiVersion >= 1.016 then
+    PageFiles[#PageFiles + 1] = { title = "Profiles", script = "profiles.lua" }
+end
+
+if apiVersion >= 1.016 then
     PageFiles[#PageFiles + 1] = { title = "PIDs 1", script = "pids1.lua" }
 end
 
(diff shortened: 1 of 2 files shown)


src/SCRIPTS/BF/pages.lua
```diff
@@ -4,6 +4,10 @@ if apiVersion >= 1.036 then
     PageFiles[#PageFiles + 1] = { title = "VTX Settings", script = "vtx.lua" }
 end
 
+if apiVersion >= 1.016 then
+    PageFiles[#PageFiles + 1] = { title = "Profiles", script = "profiles.lua" }
```
haslinghuis, line 8: Move to line 12?

klutvott123: By having it like this the pages can be shuffled around as we please. and if a page is to be removed we have to break it out anyway. I also like the readability of the current approach of having a if for each page.

haslinghuis (approved):

PR #438, opened 2022-05-31
fix shebang in bin/build.sh

sorki:
Using `env` is a more portable approach.

### DIFF
--- a/bin/build.sh
+++ b/bin/build.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 if [ -d obj ]; then
     rm -fR obj/*


haslinghuis:
Good find. But not sure if we need anything to fix.

https://stackoverflow.com/questions/21612980/why-is-usr-bin-env-bash-superior-to-bin-bash

haslinghuis:
Good find. But not sure if we need anything to fix.

https://stackoverflow.com/questions/21612980/why-is-usr-bin-env-bash-superior-to-bin-bash

sorki:
I found it by cloning the repo and trying to run `make` to test vs betaflight 4.3. My distro (NixOS) doesn't have `/bin/bash` (not even `/usr/bin/bash`) so we need to use `env`. The change is pretty harmless and should work everywhere.

PR #457, opened 2022-11-12
Trim acc

druckgott:
Added A new Page fro Trim Accelerometer


src/SCRIPTS/BF/PAGES/acc_calibrate.lua
```diff
@@ -0,0 +1,30 @@
...
+local tableSpacing = template.tableSpacing
+local sp = template.listSpacing.field
+local yMinLim = radio.yMinLimit
+local x = margin
+local y = yMinLim - lineSpacing
+local inc = { x = function(val) x = x + val return x end, y = function(val) y = y + val return y end }
+local labels = {}
+local fields = {}
+
+if apiVersion >= 1.041 then
```
haslinghuis, line 15:
```suggestion
if apiVersion >= 1.16 then
```

src/SCRIPTS/BF/PAGES/acc_calibrate.lua
```diff
@@ -0,0 +1,30 @@
...
+	labels[#labels + 1] = { t = "Trim Accelerometer" ,     x = x,          y = inc.y(lineSpacing) }
+	fields[#fields + 1] = { t = "acc trim pitch",    x = x, y = inc.y(lineSpacing), sp = x + sp, min = -300, max = 300, vals = { 1 } }
+	fields[#fields + 1] = { t = "acc trim roll",     x = x, y = inc.y(lineSpacing), sp = x + sp, min = -300, max = 300, vals = { 2 } }
```
klutvott123, line 18: Too much indent. 
The field text could also be just "Pitch" and "Roll" since the "Trim Accelerometer" header explains it all. These should also be indented. "x = x + indent" for both of these. Look at the other pages to see how we usually do it.

This won't work as it should. both pitch and roll trim are 16 bit values so for pitch it should be "vals = { 1, 2 }" and for roll "vals = { 3, 4 }". Have a look at the MSP_ACC_TRIM" message in the BF source code. 🙂

src/SCRIPTS/BF/pages.lua
```diff
@@ -56,4 +56,8 @@ if apiVersion >= 1.41 then
     PageFiles[#PageFiles + 1] = { title = "GPS PIDs", script = "gpspids.lua" }
 end
 
+if apiVersion >= 1.16 then
+    PageFiles[#PageFiles + 1] = { title = "Trim Accelerometer", script = "acc_calibrate.lua" }
```
klutvott123, line 60: maybe "acc_trim.lua" would be a more appropriate filename since we already have similarly named files to calibrate the accelerometer.

src/SCRIPTS/BF/PAGES/acc_calibrate.lua
```diff
@@ -0,0 +1,30 @@
...
+end
+
+return {
+    read        = 240, -- MSP_ACC_TRIM
+    write       = 239, -- MSP_SET_ACC_TRIM         
+    title       = "ACC",
+    reboot      = true,
+    eepromWrite = true,
+    minBytes    = 1,
```
klutvott123, line 27: For this message we need at least 4 bytes to be able to do anything.

src/SCRIPTS/BF/PAGES/acc_calibrate.lua
```diff
@@ -0,0 +1,30 @@
...
+	labels[#labels + 1] = { t = "Trim Accelerometer" ,     x = x,          y = inc.y(lineSpacing) }
+	fields[#fields + 1] = { t = "acc trim pitch",    x = x, y = inc.y(lineSpacing), sp = x + sp, min = -300, max = 300, vals = { 1 } }
+	fields[#fields + 1] = { t = "acc trim roll",     x = x, y = inc.y(lineSpacing), sp = x + sp, min = -300, max = 300, vals = { 2 } }
+end
+
+return {
+    read        = 240, -- MSP_ACC_TRIM
+    write       = 239, -- MSP_SET_ACC_TRIM         
+    title       = "ACC",
+    reboot      = true,
```
klutvott123, line 25: Is a reboot needed for the trim to take effect?

klutvott123 (changes requested):

src/SCRIPTS/BF/PAGES/acc_calibrate.lua
druckgott: I don't know if it's needed. How can I test this?

Thx for all the other informations it's my fist commit,didnt know all the thinks only test it and it works 😉.

What happens if I do Wirte only 1 instead of 1, 2 due to 16 bit?

klutvott123: "reboot = false". Then try making some changes, save the page and see if it took effect 🙂 

What happens is that the pitch field will modify the first byte of pitch trim and the roll field will modify the second byte of pitch trim. So the roll field will make unintended changes to pitch trim, and you won't be able to change roll trim at all

druckgott:
Fixed all points and tested it. Worked for me.

klutvott123:
Negative values doesn't display properly. We haven't had to deal with negative values in the bf lua scripts until now so everything is treated as unsigned. We'll have to look into that

src/SCRIPTS/BF/PAGES/acc_trim.lua
```diff
@@ -0,0 +1,27 @@
...
+local yMinLim = radio.yMinLimit
+local x = margin
+local y = yMinLim - lineSpacing
+local inc = { x = function(val) x = x + val return x end, y = function(val) y = y + val return y end }
+local labels = {}
+local fields = {}
+
+labels[#labels + 1] = { t = "Trim Accelerometer",	x = x,          y = inc.y(lineSpacing) }
+fields[#fields + 1] = { t = "Pitch",    			x = x + indent, y = inc.y(lineSpacing), sp = x + sp, min = -300, max = 300, vals = { 1, 2 } }
+fields[#fields + 1] = { t = "Roll",     			x = x + indent, y = inc.y(lineSpacing), sp = x + sp, min = -300, max = 300, vals = { 3, 4 } }
```
klutvott123, line 16: Looks like you're using tabs here. use spaces to align things.

klutvott123 (changes requested):

druckgott:
Ok yes I also so this this -1 ore <=-1, do you have an Idea where we have to change the code to get this working. Is there any way ob debuggin?

druckgott:
Ok now the problem with negativ values is fixed. I do not exactly now there the limit by switching to negativ calculation so I selected 65000 because this will never reached from -300 to 300

klutvott123:
@druckgott I think it's better if we handle this in a more generic way. See https://github.com/betaflight/betaflight-tx-lua-scripts/pull/458
With this your PR should work without the postLoad function

druckgott:
ok I removed it again and if your fix is done then it should be fine.

klutvott123:
Looks good to me. You just need to squash to one commit now 👍

druckgott:
How can I squash to one commit now?

klutvott123:
`git rebase -i HEAD~9`
you'll get a list of the last 9 commits and a bunch of actions. every commit is prepended with "pick" as default. Replace "pick" with "fixup" or "f" for all commits except the top one that you want to keep. This will leave you with one commit with all your changes.
Then `git push --force-with-lease` to push your changes

druckgott:
Mhh I cant get it working. I did all the things bevor online direct on github.com and not with git clone ....
can you fix it, next time I will directly do it with git on windows.

klutvott123:
Just make a new PR

PR #461, opened 2022-11-17
Fix nightly build

klutvott123:
I hope 😂


haslinghuis (approved):

klutvott123:
and it didn't. I see the error here https://github.com/betaflight/betaflight-tx-lua-scripts/actions/runs/3492403515/jobs/5846153329, but it doesn't tell me much. Any idea why it's not working? @blckmn ?

PR #473, opened 2023-03-05
Add level angle limit

klutvott123:
and stick sensitivity

### DIFF
--- a/src/SCRIPTS/BF/PAGES/pid_advanced.lua
+++ b/src/SCRIPTS/BF/PAGES/pid_advanced.lua
@@ -11,6 +11,14 @@ local inc = { x = function(val) x = x + val return x end, y = function(val) y = 
 local labels = {}
 local fields = {}
 
+if apiVersion >= 1.31 then
+    labels[#labels + 1] = { t = "Angle/Horizon",     x = x,          y = inc.y(lineSpacing) }
+    fields[#fields + 1] = { t = "Angle Limit",       x = x + indent, y = inc.y(lineSpacing), sp = x + sp, min = 10, max = 90, vals = { 18 } }
+    if apiVersion < 1.36 then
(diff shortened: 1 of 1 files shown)


src/SCRIPTS/BF/PAGES/pid_advanced.lua
```diff
@@ -11,6 +11,14 @@ local inc = { x = function(val) x = x + val return x end, y = function(val) y =
 local labels = {}
 local fields = {}
 
+if apiVersion >= 1.31 then
+    labels[#labels + 1] = { t = "Angle/Horizon",     x = x,          y = inc.y(lineSpacing) }
+    fields[#fields + 1] = { t = "Angle Limit",       x = x + indent, y = inc.y(lineSpacing), sp = x + sp, min = 10, max = 90, vals = { 18 } }
+    if apiVersion < 1.36 then
+        fields[#fields + 1] = { t = "Stick Sensitivity", x = x + indent, y = inc.y(lineSpacing), sp = x + sp, min = 10, max = 200, vals = { 19 } }
+    end
```
haslinghuis, line 19: We removed support for 1.41 and below.

klutvott123: Are you sure? I think it was removed in 3.2.0/1.36 https://github.com/betaflight/betaflight/blob/36c91c76f6bae1c31ad061be555beac7c163ee1b/src/main/fc/fc_msp.c#L1183

haslinghuis: No I mean we removed support for API 1.41 and below in general as we did in configurator assuming 

    if apiVersion >= 1.31 then 

would always be valid and no longer needed.

klutvott123: It's still possible to use the lua scripts with older firmware versions, so no, it won't always be valid. Conditionals like this also serve as documentation.
I don't think we should limit backwards compatibility in the lua scripts as long as it's not a maintenance nightmare. 🙂

haslinghuis (approved):

PR #475, opened 2023-03-12
Skip vtx tables download if not available

klutvott123:
Use msp error bit to determine if message is available in firmware or not. Pass msp error to the layer above so that we can take action if there's an error. 
The FC will still respond if for example MSP_VTX_CONFIG isn't available, but the msp over telemetry error bit will be set and the payload is an error code. We can use this to detect if the firmware has been compiled with VTX support, and skip vtx tables download if not.

[betaflight-tx-lua-scripts_1.7.0.zip](https://github.com/betaflight/betaflight-tx-lua-scripts/files/10951850/betaflight-tx-lua-scripts_1.7.0.zip)

### DIFF
--- a/src/SCRIPTS/BF/MSP/common.lua
+++ b/src/SCRIPTS/BF/MSP/common.lua
@@ -6,6 +6,7 @@ local MSP_STARTFLAG = bit32.lshift(1,4)
 local mspSeq = 0
 local mspRemoteSeq = 0
 local mspRxBuf = {}
+local mspRxError = false
 local mspRxSize = 0
 local mspRxCRC = 0
 local mspRxReq = 0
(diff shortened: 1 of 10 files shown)


klutvott123:
Added the message "N/A" that will replace the content of the page if it can't be loaded. Updated first post with new zip file if anyone wants to try it.

TheIsotopes:
@klutvott123 
for the osd elements page there should also be a query as to whether the firmware was compiled with osd.
lua scripts must be ended with long RTN because an error message appears.

klutvott123:
Where exactly does this error message appear? And what does it say?

MSP_OSD_CONFIG isn't wrapped in `"#IF defined(USE_OSD)` like it should have been, so it will return something even if the firmware is compiled without OSD. This means we can't detect it in the same way as for vtx and gps rescue for example. The way it's done in the firmware is very inconsistent between messages. it's annoying and we should look into it

TheIsotopes:
here the sreenshot from error:

screen-2023-03-12-230014

klutvott123:
@TheIsotopes And this is without OSD, OSD_SD and OSD_HD?

haslinghuis (approved):

TomAlperin:
Tested and working properly although I did notice that for analog, the second load is a little bit faster. Tested with my previous work around of creating the files manually and that prevents the VTX table check making the load time a bit faster for digital video systems.

klutvott123:
@TomAlperin Thanks for testing. I made it so that it creates a empty vtx table if not available. This will make it skip future download attempts
[betaflight-tx-lua-scripts_1.7.0.zip](https://github.com/betaflight/betaflight-tx-lua-scripts/files/10957679/betaflight-tx-lua-scripts_1.7.0.zip)

klutvott123:
@TheIsotopes Can you try this? https://github.com/betaflight/betaflight/pull/12513

haslinghuis (approved):

TheIsotopes (approved):

PR #480, opened 2023-05-25
Hide VTX menu when feature is unavailable

atomgomba:
This is one more step in handling the absence of VTX tables, completely hiding the option in the main menu would make scrolling a tiny bit more convenient and the user won't have to wait a few seconds for no reason just to see "N/A" when clicking on VTX menu.

Also I'm not sure if this is a good solution, I was thinking about potentially being able to hide options for other features as well (e. g. some users might have no GPS or OSD and we could hide those too, or whatever optional feature may come up in the future). Hiding the VTX menu alone would require a simpler logic, hence less new lines added.

### DIFF
--- a/src/SCRIPTS/BF/features.lua
+++ b/src/SCRIPTS/BF/features.lua
@@ -1,0 +1,5 @@
+local features = {
+    vtx = true,
+}
+
+return features
--- a/src/SCRIPTS/BF/ui.lua
+++ b/src/SCRIPTS/BF/ui.lua
@@ -82,6 +82,24 @@ local function confirm(page)
     collectgarbage()
 end
 
+local function filterAvailablePages(pageFiles)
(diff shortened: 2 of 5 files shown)


haslinghuis:
Nice - another consumer of https://github.com/betaflight/betaflight/pull/12512 :)

atomgomba:
> Pull request closed automatically as inactive.

Bad bot

atomgomba:
Please accept this during Hacktoberfest :P

Asizon:
Lol, we need the Lua master here @klutvott123

atomgomba:
just out of curiosity, why is this stalled? any objections? @klutvott123

TheIsotopes:
@atomgomba It would be nice if they could implement this for the OSD and GPS page as well. 👍

PR #496, opened 2024-01-24
Hide OSD (SD) settings

atomgomba:
Follow up PR on #495. Please merge #495 first, because it contains the parent commit.

### DIFF
--- a/src/SCRIPTS/BF/PAGES/pos_osd.lua
+++ b/src/SCRIPTS/BF/PAGES/pos_osd.lua
@@ -1,9 +1,7 @@
 local template = assert(loadScript(radio.template))()
 local margin = template.margin
-local indent = template.indent
 local lineSpacing = template.lineSpacing
 local tableSpacing = template.tableSpacing
-local sp = template.listSpacing.field
 local yMinLim = radio.yMinLimit
 local x = margin
 local y = yMinLim - lineSpacing
@@ -133,12 +131,12 @@ return {
         self.values[3] = bit32.rshift(combineValue, 8)
         return self.values
     end,
-    checkProfile = function(self, value, profileFirst, profileSecond, profileThird)
+    checkProfile = function(_, value, profileFirst, profileSecond, profileThird)
         local profiles = profileFirst + bit32.lshift(profileSecond, 1) + bit32.lshift(profileThird, 2)
         local output = bit32.replace(value, profiles, 11, 3)
         return output
     end,
-    splitVal = function(self, inputFirstVal, inputSecondVal)
+    splitVal = function(_, inputFirstVal, inputSecondVal)
         local inputVal = inputFirstVal + bit32.lshift(inputSecondVal, 8)
         local profiles = bit32.extract(inputVal, 11, 3)
         local fieldPos = bit32.extract(inputVal, 0, 11)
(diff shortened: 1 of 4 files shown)


atomgomba:
It looks like on some boards getting the OSD config may take a lot of time over CRSF

TheIsotopes:
Tested and everything seems to work.

Imo a rebase is still missing here.

atomgomba:
> Tested and everything seems to work.
> 
> Imo a rebase is still missing here.

Seemingly it doesn't *always* work for me. I tested using 4 different quads. My radio is Tango 2 with FreedomTX, what's your radio? I'm starting to feel like I'm gonna need more hardware for testing. I can get a Micro TX for my Taranis X9D+ to have one more test device.

TheIsotopes:
I use the RM Zorro with EdgeTX for FrSky and Ghost receivers.

atomgomba:
With [#13333](https://github.com/betaflight/betaflight/pull/13333) we're slowly getting to the point where firing a single MSP request to get the supported options will be enough. Waiting for multiple responses takes inconveniently lot amount of time, I'm planning to put this PR on hold until we have a better way to check build options.

klutvott123:
> With [#13333](https://github.com/betaflight/betaflight/pull/13333) we're slowly getting to the point where firing a single MSP request to get the supported options will be enough. Waiting for multiple responses takes inconveniently lot amount of time, I'm planning to put this PR on hold until we have a better way to check build options.

This is probably the best solution. I think we should put everything else on hold just to avoid ending up with multiple solutions and compatibility issues. We have three solutions to this right now, and we should pick just one. 
I'm not sure which protocol you're testing this with, but the worst case scenario is FPORT as it's super slow.

atomgomba:
> > With [#13333](https://github.com/betaflight/betaflight/pull/13333) we're slowly getting to the point where firing a single MSP request to get the supported options will be enough. Waiting for multiple responses takes inconveniently lot amount of time, I'm planning to put this PR on hold until we have a better way to check build options.
> 
> This is probably the best solution. I think we should put everything else on hold just to avoid ending up with multiple solutions and compatibility issues. We have three solutions to this right now, and we should pick just one. I'm not sure which protocol you're testing this with, but the worst case scenario is FPORT as it's super slow.

Currently I've only tested on CRSF, but on two "different" platforms, FreedomTX and EdgeTX. I happen to own an X9D+ and FrSky receivers so with a bit of soldering I can test FPORT too. Thanks for the heads up!

atomgomba:
New plan is to have a dedicated MSP command for clients using OTA transmission, please see [#13344](https://github.com/betaflight/betaflight/pull/13344)

atomgomba:
This is not stale, just waiting for [#13333](https://github.com/betaflight/betaflight/pull/13333) to be merged. I presume it's most likely to happen after 4.5 release.

haslinghuis:
@TheIsotopes pls test again :) (needs latest dev firmware)

atomgomba:
> @TheIsotopes pls test again :) (needs latest dev firmware)

well, it's only needed to test the happy path. by design the menu will load either way ;) btw I also tested this, but more tests the better. testing against MSP version <= 1.46 is a real scenario as well. so far I haven't tested against BF 4.4, but I'm planning to do so in the weekend

TheIsotopes:
@atomgomba @haslinghuis
tested on my Zorro, everything seems to be working correctly for me. 👍

haslinghuis (approved):

atomgomba:
tested on 4.4.3, works as expected

Issue #505, opened 2024-04-09
Req: Battery voltage warnings setup via LUA from radio


neprsitady:
Hey team,

would it be possible to add the Battery voltage warning and minimum values to the LUA scripts as well? 

Use case: Currently have the warnings set for li ion batteries but fly with lipos as well and cant change these without computer. (Profile dependent battery settin would be great in BF, but I see it could bring troubles to begginers)

PR #515, opened 2024-12-29
Add battery voltage warnings setup via LUA from radio

vishwamartur:
Related to #505

Add battery voltage warning and minimum values configuration via LUA scripts from the radio.

* **New LUA Script**: Add `src/SCRIPTS/BF/battery_voltage.lua` to handle battery voltage warning and minimum values configuration.
  * Define functions to read and set battery voltage warning values for different battery types (Li-ion, LiPo).
  * Use MSP commands to communicate with the flight controller.
* **Pages Update**: Modify `src/SCRIPTS/BF/pages.lua` to include a new page entry for battery settings configuration.
  * Add a new page entry with the title "Battery Settings" and script "battery_voltage.lua".
* **UI Update**: Modify `src/SCRIPTS/BF/ui.lua` to include the new battery settings page in the UI menu.
  * Add the new battery settings page to the UI menu.

### DIFF
--- a/src/SCRIPTS/BF/battery_voltage.lua
+++ b/src/SCRIPTS/BF/battery_voltage.lua
@@ -1,0 +1,46 @@
+local MSP_SET_BATTERY_CONFIG = 210
+local MSP_BATTERY_CONFIG = 211
+
+local batteryConfig = {
+    warningVoltage = 0,
+    minVoltage = 0,
+    batteryType = 0
+}
+
+local function processMspReply(cmd, payload, err)
+    if cmd == MSP_BATTERY_CONFIG and not err then
+        batteryConfig.warningVoltage = payload[1]
+        batteryConfig.minVoltage = payload[2]
+        batteryConfig.batteryType = payload[3]
+    end
+end
+
+local function getBatteryConfig()
+    protocol.mspRead(MSP_BATTERY_CONFIG)
+    mspProcessTxQ()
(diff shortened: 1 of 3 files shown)


klutvott123:
Is there a reason why this can't just be a regular page like? 🙂

Forceu:
I wanted to built something like this on my own, but saw your PR - thank you and I hope it gets merged at some point.

After compiling the script, I can however not open the battery page, when selecting it on a monochrome display, I get the error that `battery_voltage.lua` was not found. After moving the file to the `SCRIPTS/BF/PAGES` folder (currently it is in the `SCRIPTS/BF` folder), it finds the file, but then I get the error

> Syntax error
> ?:0
> attempt to index field 'fields' (a nil value).

Looking through the other page files, all start with 

```
local template = assert(loadScript(radio.template))()
local margin = template.margin
local indent = template.indent
local lineSpacing = template.lineSpacing
local tableSpacing = template.tableSpacing
local sp = template.listSpacing.field
local yMinLim = radio.yMinLimit
local x = margin
local y = yMinLim - lineSpacing
local inc = { x = function(val) x = x + val return x end, y = function(val) y = y + val return y end }
local labels = {}
local fields = {}
```

and have a different syntax. So I think it needs to be modified to work, before it can be merged.

klutvott123:
This is not the correct way to do it. Closing

PR #542, opened 2026-07-05
added battery presets for changing chemistry types

tmacinc:
Added Presets to the Battery page for easy switching of types.

The new betaflight battery profiles and switch option might make this unnecessary. But it's a nice option to see and confirm the settings before you arm as part of your preflight if you're a user that switches types frequently.

## Summary by CodeRabbit

* **New Features**
  * Added a battery type preset selector with common options like Custom, LiPo, LiIon, and LiHV.
  * The battery settings screen can now automatically match the preset based on existing cell voltage values.
  * Changing the battery type now updates the related voltage fields to keep settings consistent.

### DIFF
--- a/src/SCRIPTS/BF/PAGES/battery.lua
+++ b/src/SCRIPTS/BF/PAGES/battery.lua
@@ -11,10 +11,31 @@ local inc = { x = function(val) x = x + val return x end, y = function(val) y = 
 local labels = {}
 local fields = {}
 
+-- Preset voltage values (in volts, matching scale = 100)
+local batteryPresets = {
+    [0] = { min = nil,  max = nil,   warn = nil  }, -- Custom (no preset applied)
+    [1] = { min = 3.30, max = 4.20,  warn = 3.50 }, -- LiPo
+    [2] = { min = 2.80, max = 4.20,  warn = 3.00 }, -- LiIon
+    [3] = { min = 3.30, max = 4.35,  warn = 3.50 }, -- LiHV
+}
+
 labels[#labels + 1] = { t = "Voltage Settings",     x = x, y = inc.y(lineSpacing) }
-fields[#fields + 1] = { t = "Minimum Cell",         x = x + indent, y = inc.y(lineSpacing), sp = x + sp, min = 0, max = 500, vals = { 8, 9 }, scale = 100 }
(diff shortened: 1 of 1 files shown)


haslinghuis:
@YarosMallorca does this complement #544 ?

YarosMallorca:
@haslinghuis Seems like this is a workaround that works on older versions, as far as I understand, this just tweaks the values of the current profile instead of switching between the newly-introduced battery profiles (like my PR does).
#544 is only available on 26.06 or higher, because it's tied to the firmware-side implementation of battery profiles. 

So answering your question: yes, you could say it complements my PR, but also it's just a different feature that solves a very similar problem.

Keep in mind that this PR hardcodes the "presets" in [battery.lua](https://github.com/betaflight/betaflight-tx-lua-scripts/pull/542/changes#diff-fe9da780ae7d3c3a3750a9972deb20e0cd27010de91e2dc45970ecf3bf369148R15-R20). 

Taking into account that Betaflight 26.06 already supports native battery profiles, is this workaround even needed?

tmacinc:
> @haslinghuis Seems like this is a workaround that works on older versions, as far as I understand, this just tweaks the values of the current profile instead of switching between the newly-introduced battery profiles (like my PR does). #544 is only available on 26.06 or higher, because it's tied to the firmware-side implementation of battery profiles.
> 
> So answering your question: yes, you could say it complements my PR, but also it's just a different feature that solves a very similar problem.
> 
> Keep in mind that this PR hardcodes the "presets" in [battery.lua](https://github.com/betaflight/betaflight-tx-lua-scripts/pull/542/changes#diff-fe9da780ae7d3c3a3750a9972deb20e0cd27010de91e2dc45970ecf3bf369148R15-R20).
> 
> Taking into account that Betaflight 26.06 already supports native battery profiles, is this workaround even needed?

This is a good description of how this works. I built it in my fork and was using it as a way to easily change battery types in the field. 

I'm not sure if its still a useful addition since profiles were added, and I havent checked if there are any negatives with the newest version as Ive been really busy. 

If you're able to select profiles and edit their values from the .Lua this is probably a legacy addition that isn't needed. It was a great feature before presets were added.

YarosMallorca:
> If you're able to select profiles and edit their values from the .Lua this is probably a legacy addition that isn't needed. It was a great feature before presets were added.

Correct, now you can switch battery profiles from the Profile screen and change their values from Power/Battery tab, similar to how rateprofiles work.

tmacinc:
Closed as this is a legacy feature.

